Issue #6, opened 2020-06-19
Provide size estimation (codomain cardinality) in Arbitraries


hugoferreira:
Providing a codomain size is essential for two things (at least):

a. Knowing how much of the domain we have explored during sample;
b. Early stop of a sampling process if the codomain is estimated to be zero;
c. Providing a confidence level for universal and existential assertions.

Co-domain size is known to be exact in some primitive Arbitraries. But composition of arbitraries, especially through `map`, `filter`, and `chain`, might introduce uncertainty.

The roadmap for this issue is to provide accurate `size()` in:

  - [ ] Mapped Arbitraries
  - [x] Filtered Arbitraries
  - [ ] Unique Arbitraries
    - [ ] Mathematical treatment of Unique Arbitrary vs sampling without replacement 
  - [ ] Primitive Arbitraries
    - [x] Integer
    - [ ] Real
    - [x] Boolean
    - [x] String
    - [ ] Collection (Array)
    - [ ] Composite (Union)

hugoferreira:
## Related Issues

- #457 - Exhaustive generation for small arbitraries (uses `size()` to determine when enumeration is feasible)
- #8 - Accurate size estimation for mapped arbitraries
- #9 - Filter arbitrary size estimation improvements

PR #48, opened 2020-06-25
Add confidence values for small integers

ruippeixotog:
Depends on #47. Comments and explanations in the code.

### DIFF
--- a/test/confidence.test.ts
+++ b/test/confidence.test.ts
@@ -3,7 +3,7 @@ import { expect } from 'chai'
 import * as fc from '../src'
 import { Arbitrary } from '../src/arbitraries'
 
-describe('Confidence tests', () => {
+describe('Confidence tests', function () {
 
   /**
    * The one true test to evaluate the confidence returned by an arbitrary, given the population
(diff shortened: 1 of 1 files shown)


hugoferreira:
## Closing as Stale - Confidence Calculation Documented

This PR from **June 2020** (5 years old, 349 commits behind master) implemented confidence calculation for small integers with a formal mathematical specification.

### Formal Definition of Confidence

From the test specification:
> *"The confidence provided by an arbitrary is the probability that, when asked for a sample of a given size, the returned sample will result in the correct conclusion about the truthness of the property."*

### Testing Approach

Exhaustive property enumeration for verification:
```typescript
// Given a population, returns the set of all possible distinct properties
const allPropsFor = <A>(pop: A[]): ((a: A) => boolean)[] => {
  const revIndex = new Map(pop.map((a, i) => [a, i]))
  const arr: ((a: A) => boolean)[] = []
  for (let bitset = 0; bitset < 2 ** pop.length; bitset++) {
    arr.push(a => (bitset & (1 << revIndex.get(a)!)) !== 0)
  }
  return arr
}
```

### Connection to Current Work

This formal specification is valuable for:
- #413 - Research: Statistical Ergonomics (defines what "confidence" means)
- #418 - Confidence-based termination (needs confidence calculation)
- #6 - Size estimation (confidence depends on domain size)

### Why Not Merge?

1. **Depends on #47** - Which changes `sample()` return type
2. **Massive divergence** - 349 commits behind master
3. **Computationally expensive** - As noted in comments, exhaustive property enumeration is O(2^n)
4. **Superseded approach** - Current statistical ergonomics uses different architecture

Reference: https://github.com/fluent-check/fluent-check/tree/add-confidence-for-small-integers

Issue #115, opened 2021-02-24
Abstract the test case generation process


RuiGuedes:
Currently, there is no support for different types of generating test cases, which are generated at once. That is only possible because the test case generation process is entirely random, which means that the generated inputs do not influence each other.

Since new types of generating test cases that consider the program's feedback to generate test cases are intended to be implemented, such a process cannot be done in a single step.

Therefore the idea here is to create an interface/abstract class (let's call it Guidance for now) that allows other classes to implement/extend it. The default Guidance will be a random one (let's say RandomGuidance). In this way, it will become possible to alter the current FluentCheck architecture while maintaining the same type of generation that it actually supports.

RuiGuedes:
Solved in #117

Issue #18, opened 2020-06-20
Should shrink perform an exhaustive search by default?


hugoferreira:
The following test fails:

```typescript
expect(new FluentCheck()
      .exists('a', fc.integer(0, 1000000))
      .then(({ a }) => a % 11 === 0 && a > 90000 && a < 90010)
      .check()
).to.deep.include({ satisfiable: true, example: { a: 90002 } })
```

... because it is statistically unlikely for the sampler to generate (and then shrink) the number we are expecting, as it cannot analyse our filter condition. There are several paths we can take from here:

1. Choose shrink to work on a best-effort basis;
1. Choose exhaustive search to be possible and parameterised by the user;
1. Choose exhaustive search to be selected based on heuristics provided by the `size()` estimators.

Pros/cons and voting, please :)

PR #117, opened 2021-02-25
Abstracting the test case generation process

RuiGuedes:
This PR addresses issue #115. Additional information will be added here as the development advances.

### DIFF
--- a/src/strategies/RandomStrategy.ts
+++ b/src/strategies/RandomStrategy.ts
@@ -1,0 +1,40 @@
+import {Strategy} from './Strategy'
+import {Arbitrary} from '../arbitraries'
+
+export class RandomStrategy extends Strategy {
+
+  /**
+   * Number of samples generated
+   */
+  private numGenSamples = 0
+
+  constructor(public readonly maxGenSamples) {
+    super()
+  }
+  /**
+   * NOTE -> For this particular strategy this function only
+   * generates samples to a certain point specified by maxGenSamples.
+   * Others factors (e.g. time) can be used to determine when the
+   * generation process should be stoped.
+   */
+  hasInput() {
+    return this.numGenSamples <= this.maxGenSamples
+  }
+
+  getInput<A>(a: Arbitrary<A>) {
+    return a.pick()
+  }
+
+  /**
+   * Handles the result of running a test with a certain test case.
+   * When called this function marks the end of one iteration in the
+   * test case generation process. Therefore the numGenSamples incresases
+   * by one each time this function is called.
+   *
+   * NOTE -> This function can also include some perform additional operations
+   * like keeping a list of generated inputs and save them to a file.
+   */
+  handleResult(res: Boolean) {
+    this.numGenSamples++
+  }
+}
--- a/src/strategies/Strategy.ts
+++ b/src/strategies/Strategy.ts
@@ -1,0 +1,18 @@
+import {Arbitrary, FluentPick} from '../arbitraries'
+
+export abstract class Strategy {
+  /**
+   * Determines whether there are still inputs to be generated or not
+   */
+  abstract hasInput(): Boolean
+
+  /**
+   * Generates a new input
+   */
+  abstract getInput<A>(a: Arbitrary<A>): FluentPick<any> | undefined
+
+  /**
+   * Handles the result of running a particular test case
+   */
+  abstract handleResult(res: Boolean): void
+}


src/strategies/FluentBiasedRandomStrategy.ts
```diff
@@ -0,0 +1,29 @@
+import {Arbitrary} from '../arbitraries'
+import {FluentResult} from '../FluentCheck'
+import {FluentRandomStrategy} from './FluentRandomStrategy'
+
+export class FluentBiasedRandomStrategy extends FluentRandomStrategy {
```
arestivo, line 5: I would prefer composition over inheritance here. For example, a FluentBiasedDecorator(FluentRandomStrategy) would ask the base strategy for inputs and add its own biased values.

RuiGuedes: I thought about that because at first I even thought of making the shrinking process optional.  However, since I agreed with what @ruippeixotog said in #18 I saw fit only to consider the possibility of having bias or not, and therefore inheritance seemed suitable here.

> I always thought of shrinking as a best-effort, nice-to-have feature of PBT.

Nevertheless, I understand your point of view and therefore I will consider your suggestion and change my implementation.

RuiGuedes:
I have marked this PR as ready for review but there are still some things at the test level that I need to do. In fact, I think I found some interesting cases but I need to investigate a little more. However, in the meanwhile @arestivo and @hugoferreira feel free to check what I did and to give me feedback about it.

RuiGuedes:
I have marked this PR as ready for review but there are still some things at the test level that I need to do. In fact, I think I found some interesting cases but I need to investigate a little more. However, in the meanwhile @arestivo and @hugoferreira feel free to check what I did and to give me feedback about it.

RuiGuedes:
I have just finished all the code related to this PR. However, since this PR depends on PR #120 I still need to perform additional checking at the test level before merging this PR.

Btw before merging this PR I would like to provide a better description than the currently used one so that it becomes easier for me to gather all the information about what I have done when writing the dissertation.

hugoferreira (dismissed):
We need to rebase this PR after merging #131

src/strategies/FluentStrategy.ts
```diff
@@ -1,65 +1,131 @@
...
-   * Adds an arbitrary to the arbitraries record.
-   */
-  abstract addArbitrary<K extends string, A>(name: K, dedup: Arbitrary<A>)
+  setCurrArbitraryDedupHook() {}
+
+  setCurrArbitraryCacheHook() {}
+
+  isDedupable() { return false }
+
+  isBiased() { return false }
```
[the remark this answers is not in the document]
RuiGuedes: That version you commented is outdated

test/integers.test.ts
```diff
@@ -1,10 +1,12 @@
 import * as fc from '../src/index'
+import * as Strategies from '../src/strategies/index'
 import {it} from 'mocha'
 import {expect} from 'chai'
 
 describe('Integer tests', () => {
   it('finds there is a number in the -10, 10 range with inverse and shrink it to 0', () => {
     expect(fc.scenario()
+      .config(new Strategies.RandomCachedStrategy()) // NOTE - This test still passes with a shrinkingless strategy
```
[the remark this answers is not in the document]
RuiGuedes: Yeah, you are right. I marked the test with that warning for further investigation. However, according to the test itself, I think we should change the test description and also remove the provided `example` since it makes more sense to only use the `example` field when shrinking is part of the test.

src/FluentCheck.ts
```diff
@@ -162,24 +167,18 @@ abstract class FluentCheckQuantifier<K extends string, A, Rec extends ParentRec
...
-    const collection = depth === 0 ?
-      this.cache :
-      (partial !== undefined ?
-        this.a.shrink(partial.example[this.name]).sampleUniqueWithBias(this.configuration.shrinkSize) :
-        [])
+    this.strategy.configArbitrary(this.name, partial, depth)
 
-    for (const tp of collection) {
-      testCase[this.name] = tp
+    while (this.strategy.hasInput(this.name)) {
```
arestivo, line 172: Could this be designed has an iterator so that we can use a `for ... of```?

RuiGuedes: Yes, it could. In fact, it is basically what is happening behind the curtains. Do you have a reason for preferring an explicit `for`?

src/strategies/FluentStrategyMixins.ts
```diff
@@ -0,0 +1,55 @@
...
+  return class extends Base {
+    shrink<K extends string>(arbitraryName: K, partial: FluentResult | undefined, depth: number) {
+      const shrinkedArbitrary = this.arbitraries[arbitraryName].arbitrary.shrink(partial!.example[arbitraryName])
+      this.arbitraries[arbitraryName].collection = this.buildArbitraryCollection(shrinkedArbitrary,
+        this.configuration.shrinkSize)
+    }
+  }
+}
+
+export function Dedupable<TBase extends MixinStrategy>(Base: TBase) {
```
arestivo, line 33: Is this still relevant?

src/strategies/FluentStrategyMixins.ts
```diff
@@ -0,0 +1,55 @@
...
+  return class extends Base {
+    isDedupable() {
+      return true
+    }
+  }
+}
+
+export function Cacheable<TBase extends MixinStrategy>(Base: TBase) {
```
arestivo, line 41: Semantically speaking:

Not sure if the verb should be cacheable. A strategy is not cacheable. Neither are the values. We are caching the values as part of the strategy, which is different.

src/strategies/index.ts
```diff
@@ -0,0 +1,10 @@
+import {FluentStrategy} from './FluentStrategy'
+import {Random, Dedupable, Biased, Cacheable, Shrinkable} from './FluentStrategyMixins'
+
+export const RandomCachedStrategy = Cacheable(Dedupable(Random(FluentStrategy)))
+
+export const RandomCachedStrategyWithShrinking = Shrinkable(Cacheable(Dedupable(Random(FluentStrategy))))
```
arestivo, line 6: Here, you prove what I said before. It's a cached strategy, not a cacheable strategy...

test/arbitrary.test.ts
```diff
@@ -1,10 +1,12 @@
 import * as fc from '../src/index'
+import * as Strategies from '../src/strategies/index'
 import {it} from 'mocha'
 import {expect} from 'chai'
 
 describe('Arbitrary tests', () => {
   it('should return has many numbers has asked', () => {
     expect(fc.scenario()
+      .config(new Strategies.RandomCachedStrategy())
```
arestivo, line 9: Isn't there a default strategy?

src/strategies/FluentStrategyMixins.ts
hugoferreira: I agree. `Cached` seems more appropriate

src/strategies/FluentStrategyMixins.ts
RuiGuedes: It is in terms of deciding whether sampling must be unique or not. Maybe we should rename this for something else. Do you have any suggestions?

src/strategies/index.ts
RuiGuedes: Ok I will rename Chacheable to Cached

test/arbitrary.test.ts
RuiGuedes: No. The default one is the BiasedRandomChachedStrategyWithShrinking.

src/strategies/FluentStrategy.ts
```diff
@@ -0,0 +1,104 @@
+import {Arbitrary, FluentPick} from '../arbitraries'
+import {FluentConfig, FluentResult} from '../FluentCheck'
+import {StrategyArbitraries} from './types'
+
+export interface FluentStrategyInterface {
+  hasInput: <K extends string>(arbitraryName: K) => boolean
+  getInput: <K extends string, A>(arbitraryName: K) => FluentPick<A>
+  handleResult: () => void
+}
```
hugoferreira, line 9: Why do you need the K generic if it's a string? Is there something that actually extends string?

hugoferreira: Also, `Iterator` in typescript is defined as:

```typescript
interface Iterator<T> {
    next(value?: any): IteratorResult<T>;
    return?(value?: any): IteratorResult<T>;
    throw?(e?: any): IteratorResult<T>;
}

interface IteratorResult<T> {
    done: boolean;
    value: T;
}
```

Does it make sense for `FluentStrategyInterface` to extend an `Iterator` of `FluentPick`'s? Food for thought.

RuiGuedes:
@hugoferreira  @arestivo Another point for discussion. Currently, I am assuming in all strategies defined on the `index.ts` file inside the `strategies` folder that the use of unique during sampling is made by default. What should we do? Assume it by default or, for instance, rename the strategy `BiasedRandomCachedStrategyWithShrinking` to `BiasedRandomCachedStrategyWithShrinkingAndUniqueSampling` (which is not a short name at all xD)

src/strategies/FluentStrategy.ts
```diff
@@ -0,0 +1,104 @@
...
+   */
+  constructor(public readonly configuration: FluentConfig = {sampleSize: 1000, shrinkSize: 500}) {}
+
+  /**
+   * Allows the configuration of a given strategy
+   */
+  config(config: FluentConfig) {
+    for (const k in config) this.configuration[k] = config[k]
+    return this
+  }
```
hugoferreira, line 29: Maybe:

```typescript
config(config: Partial<FluentConfig>) {
    this.configuration = {...this.configuration, ...config}
    return this
}
```

So that config can be called with just one parameter, for example.

src/strategies/FluentStrategy.ts
```diff
@@ -0,0 +1,104 @@
...
+   */
+  shrink<K extends string>(_name: K, _partial: FluentResult | undefined, _depth: number) {}
+
+  /**
+   * Determines whether there are more inputs to be used for test case generation purposes. This function can use
+   * several factors (e.g. input size, time) to determine whether the generation process should be stoped or not.
+   *
+   * Returns true if there are still more inputs to be used; otherwise it returns false.
+   */
+  hasInput<K extends string>(arbitraryName: K): boolean {
```
hugoferreira, line 84: I believe this exists because this class should not be instantiated directly, right? Can't we tag it as `abstract` instead of having these `throws`?

test/arbitrary.test.ts
```diff
@@ -14,6 +16,7 @@ describe('Arbitrary tests', () => {
 
   it('should return values in the specified range', () => {
     expect(fc.scenario()
+      .config(new Strategies.RandomCachedStrategy())
```
hugoferreira, line 19: ~Pretty please for a default strategy xD~. Just realised _there is_ a default strategy, which is `BiasedRandomChachedStrategyWithShrinking`. I don't enjoy this code duplication. Can't you use something like `beforeEach` from mocha to initialize this recurrent scenario?

test/integers.test.ts
```diff
@@ -65,12 +75,4 @@ describe('Integer tests', () => {
...
     ).to.deep.include({ satisfiable: true, example: { a: 90002 } }) */
   })
-
-  it('finds that adding 1000 makes any number larger and shrinks the example', () => {
-    expect(fc.scenario()
-      .exists('a', fc.integer())
-      .then(({a}) => a + 1000 > a)
-      .check()
-    ).to.deep.include({satisfiable: true, example: {a: 0}})
-  })
```
hugoferreira, line 75: Any particular reason why this test is commented out?

hugoferreira (commented):
> `BiasedRandomCachedStrategyWithShrinkingAndUniqueSampling`

LOL'ed!

src/strategies/FluentStrategy.ts
RuiGuedes: The use of K was to maintain coherence with the code already developed by you guys, which also uses `K extends String`

src/strategies/FluentStrategy.ts
RuiGuedes: Typescript mixins do not work with abstract classes.

test/arbitrary.test.ts
RuiGuedes: Yeah, I was not liking it too. I will investigate this.

src/strategies/FluentStrategy.ts
hugoferreira: > The use of K was to maintain coherence with the code already developed by you guys, which also uses `K extends String`

We do that because we want to use 'string''s as singleton types in 'record''s. I'm not sure if it is the case here, but I'll check it out later, that's ok.

test/integers.test.ts
RuiGuedes: It is not commented out to me

hugoferreira: Sorry, I meant deleted xD

src/FluentCheck.ts
arestivo: It's already an iterator, why not make it explicit?

src/strategies/FluentStrategyMixins.ts
arestivo: I was asking because I thought sampling should always be unique. But I remembered that we had mentioned that in some cases, removing uniques might be too expensive (e.g., space-wise).

test/arbitrary.test.ts
arestivo: We should have tests specifically for each strategy.

src/strategies/index.ts
```diff
@@ -0,0 +1,10 @@
+import {FluentStrategy} from './FluentStrategy'
+import {Random, Dedupable, Biased, Cached, Shrinkable} from './FluentStrategyMixins'
+
+export const RandomCachedStrategy = Cached(Dedupable(Random(FluentStrategy)))
+
+export const RandomCachedStrategyWithShrinking = Shrinkable(Cached(Dedupable(Random(FluentStrategy))))
+
+export const BiasedRandomCachedStrategy = Cached(Biased(Dedupable(Random(FluentStrategy))))
+
+export const BiasedRandomCachedStrategyWithShrinking = Shrinkable(Cached(Biased(Dedupable(Random(FluentStrategy)))))
```
arestivo, line 10: These names and way of composing strategies is so bad. Please, at least, open an issue to transform this into a fluent interface.

```
fc.scenario().withBias().usingCache().withRandomSampling().
  forall(...
```

src/FluentCheck.ts
RuiGuedes: I don't know, it just seemed more clearer to me this way.

src/strategies/FluentStrategyMixins.ts
RuiGuedes: Oh ok, I misunderstood. Can I consider this conversation finished then?

test/arbitrary.test.ts
RuiGuedes: Yeah, we can do that. However, notice that the strategy can influence test performance. As such for every test case where shrinking is not needed we should use a _shrinkingless_ strategy to avoid additional computation no?

src/strategies/FluentStrategy.ts
RuiGuedes: Ok, feel free to check that out and let me know what is your final suggestion?

src/strategies/FluentStrategy.ts
RuiGuedes: That could be a nice implementation even for future types of configurations. I will change that!

test/integers.test.ts
RuiGuedes: It was also not deleted xD

hugoferreira: Nevermind, it moved place... stupid diff...

src/strategies/index.ts
RuiGuedes: No need, I can make the changes in this PR.

RuiGuedes: The way of composing strategies is how it is supposed to be when using mixins. I understand your point of view that it may not be the clearest thing ever. However, even providing your suggested interface on the backend the composition would be the same. 

The question here is: do we want users to be able to configure strategies or to select strategies? I was thinking about implementing the latter since the strategy composition of the former involves knowing the logic and behavior behind the mixins.

arestivo: FluentCheck is called that way because it provides a Fluent interface.

And I think developers are grown enough to read some documentation. Just give them decent defaults, and provide some examples.

src/FluentCheck.ts
arestivo: You can still use it this way, but being an iterator allows to be used with a `for ... of`.

src/strategies/FluentStrategyMixins.ts
arestivo: Yes

test/arbitrary.test.ts
arestivo: Let the current tests unchaged. They should still work and serve as a good regression testing suite.

Create a new test suite for strategies.

src/FluentCheck.ts
RuiGuedes: I agree but for future strategies where you generate one input at a time maybe it makes more sense the way I implemented I guess.

src/strategies/index.ts
RuiGuedes: Ok, I will change the approach taken to be more fluent then and to be similar to what you have specified.

test/arbitrary.test.ts
RuiGuedes: According to @arestivo suggestions, I will create a test suite specifically for strategies and also change the way strategies are configured. Therefore, what is mentioned here is no longer a problem.

src/strategies/index.ts
RuiGuedes: Already made the required changes.

test/arbitrary.test.ts
RuiGuedes: Instead of doing the proposed changes on this PR, I created an issue #144 to be picked later.

arestivo (approved):

PR #120, opened 2021-03-04
Fixing sampling with bias duplication problem for unique arbitraries

RuiGuedes:
After developing the code for PR #117 and while testing it to make sure it was working I found that some test cases' descriptions did not match what was happening behind the curtains. The major problem identified was that the shrinking process (of `ArbitraryInteger` to be specific) was not able to shrink a number to its neutral value, `0`. 

Therefore, to enhance the shrinking process I changed the conditional returning values of the `shrink()` function to return `fc.constant(upper|lower)` instead of returning `NoArbitrary`. Even though such modification was expected to work correctly for all test cases, there is one test case which now is not passing as it now enters in an infinite loop.

The debugging process led to the identification of another problem that is the duplication of inputs for unique arbitraries. This happens because `sampleWithBias()` first retrieves the corner cases and generates a sample of inputs (without duplication) and only then combines both inputs. Well, due to this sequence of events there is the possibility of combining data that together leads to the existence of duplicates.


RuiGuedes:
All the errors that led to this PR creation were successfully fixed but a new error was introduced. The error relies on the shrinking process of mapped tupples which is represented by the test `should allow shrinking of mapped tupples` (now commented). I would ask guys to review the PR and let me know what you think.

test/arbitrary.test.ts
```diff
@@ -53,11 +53,15 @@ describe('Arbitrary tests', () => {
...
-      .check()).to.deep.include({satisfiable: true, example: {point: [202, '_aaa']}})
+    // FIXME: After changing the run() cycle of universals and existentials, which nows involves the update of the
+    // this.dedup variable, this test does not pass since FluentCheck is not able to shrink result to the example
+    // provided.
+
+    // expect(fc.scenario()
+    //   .exists('point', fc.tuple(
+    //     fc.integer(50, 1000).filter(x => x > 100),
+    //     fc.string(1, 10, 'a').filter(x => x.length > 2)).map(([a, b]) => [a * 2, '_'.concat(b)]))
+    //   .check()).to.deep.include({satisfiable: true, example: {point: [202, '_aaa']}})
```
hugoferreira, line 64: I suggest we don't comment tests that are failing, without first discussing the nature of the failure and deciding if the test is erroneous or if it still makes sense. I'm ok with having a _dev_ branch with tests failing, so that we are aware a regression exists before merging them into _master_.

src/arbitraries/UniqueArbitrary.ts
```diff
@@ -25,6 +28,15 @@ export class UniqueArbitrary<A> extends WrappedArbitrary<A> {
...
   }
 
+  sampleWithBias(sampleSize = 10): FluentPick<A>[] {
+    const cornerCases = this.cornerCases()
+
+    if (sampleSize <= cornerCases.length)
+      return this.sample(sampleSize)
+
+    return this.sample(sampleSize, cornerCases)
+  }
```
hugoferreira, line 38: I'm not following the rationale here. It seems we have a method called `sampleWithBias` that only actually _samples with bias_ when the search space is larger than the biased space. Otherwise it will ignore the bias. Am I correct?

hugoferreira (changes requested):
I suggest we don't comment the failing test. I would also like the opinion on the now defined behaviour of `sampleWithBias` in the case of `UniqueArbitraries` from @arestivo and/or @ruippeixotog.

src/arbitraries/UniqueArbitrary.ts
hugoferreira: I believe this to be because _during_ the shrinking process we don't want to keep testing the corner cases over and over, particularly in situations where they've been tested before. For small enough `sampleSizes`, the corner cases can occupy a significant portion of a biased sample. This makes me wonder about a few things:

a. Currently our strategy is to assign a 100% probability of sampling a value in the bias pool; this reminds me of a discussion we had long ago about bias being represented instead by a PMF in the Arbitrary;
b. With this new code, the probability of sampling such value drops from 100% to its true probability when the `sampleSize` is lower than the bias pool; even if we have tested it before;
c. Maybe we could have a more generic mechanism of know if a particular test-case has already been generated and discard it.

test/arbitrary.test.ts
hugoferreira: Well, I've just realised this is to be merged with master xD

src/arbitraries/UniqueArbitrary.ts
RuiGuedes: I have read what you said and I have some questions:

a. What PMF means? Also, what is the difference between this sampleWithBias() function and the one that you have developed that led you to consider that this is a new strategy?
c. What do you mean by test-case? Sampling is done for each arbitrary. Do you mean knowing the values generated for each Arbitrary or for the combination of Arbitraries?

test/arbitrary.test.ts
RuiGuedes: Ok but I uncommented the test case so you can immediately reproduce the failing test case. It needs to be commented before merging.

src/arbitraries/UniqueArbitrary.ts
hugoferreira: This appears as new code in the diff; I didn't realised it was existing code xD Anyway:

a. [Probability Mass Function](https://en.wikipedia.org/wiki/Probability_mass_function); something akin to that could be a more generic way of introducing bias during sampling. We should maybe move this discussion/idea into an issue;
b. Consider that a `UniqueArbitrary` exists so that we don't test duplicate values; but during the shrinking process, new arbitraries are spawned, and so the "memory" of the values that were already generated is lost. But that doesn't mean that we should always discard previously tested values from a "single" arbitrary, because they might contribute to a new "test-case" when considering the "combination" of arbitraries. What I'm wondering is if that shouldn't be accounted for. Again, I would move this discussion into a new Issue so that we can think about it in the future.

test/arbitrary.test.ts
hugoferreira: @arestivo please provide advice :p

src/arbitraries/UniqueArbitrary.ts
RuiGuedes: a. I agree with you and I think such discussion despite making sense to be made is out of the scope of what this PR is really about.
b. Yeah I understand what you mean and I also agree we should move such discussion to a new issue. In fact, storing the values generated for each arbitrary is something that I already considered while developing the code for PR #117.

test/arbitrary.test.ts
arestivo: I think we should first understand why the test is failing. The test itself seems to be correct, so I'm against commenting and ignoring the test just to be able to merge this PR.

src/arbitraries/UniqueArbitrary.ts
arestivo: > c. Maybe we could have a more generic mechanism of know if a particular test-case has already been generated and discard it.

Ok, as long as we are not discarding a single value from an Arbitrary but a set of values from a test case. The same value might give different results depending on the values of other arbitraries.

test/arbitrary.test.ts
RuiGuedes: @arestivo I agree with you but I think the changes done in this PR make more sense than holding the PR just for one test. Without the introduced changes, there are several tests that according to their description seem to be correct but they are not. Therefore the current version of "master" is giving false confidence.

Still, I agree with you. However, the failing test is not trivial at all to easily debug. I will try to spend a little bit more time trying to understand what is wrong.

src/arbitraries/UniqueArbitrary.ts
RuiGuedes: Yeah, keeping track of the values used by an arbitrary as well as the values that were combined with it has its complexity. Therefore the need for an issue for this problem.

arestivo: But the issue is due to this PR. So we should fix it in this PR.

hugoferreira (approved):

ruippeixotog:
Regarding API: I slightly favor `sampleForBias` keeping the uniqueness of an arbitrary, but I'm ok with any behavior for `sampleForBias` as long as the function's contract is documented and applied consistently on all other arbitraries. The concept of a `UniqueArbitrary` in itself it's strange and uncommon on the PBT libraries I used before, so we're free to choose what we want.

Users define properties in terms of existential and universal (`exists` and `forall`) statements over domains (`integer`, `string`, `const`...). On the language of propositions it makes no difference if we test the same set of inputs once or more than once, as the end result would be the same (assuming purely functional code). IMO `unique()` shouldn't exist at all as a public API as it's simply an implementation/optimization aspect. FluentCheck users' concerns are to specify correct properties in a clear manner, while FluentCheck's responsibility is to try to falsify properties as efficiently as possible.

arestivo:
> On the language of propositions it makes no difference if we test the same set of inputs once or more than once, as the end result would be the same (assuming purely functional code).

What if I want to do something like this:

```  
fc.scenario()
    .forall('a', fc.array(fc.integer(10, 20).unique(), 5, 10))
    .then(({a}) => a.length === new Set(a).size)
    .check()
```

Not that the above code works...

ruippeixotog:
That's a good example! I consider the distinctness in the elements of the array to be a property of the array and not a property of its elements. That makes sense from a set theory standpoint (in a type system you'd model an array of distinct integers as a `DistinctArray<Int>` or a `Set<Int>` rather than an `Array<DistinctInt>`) and implementation-wise it makes API contracts clearer (e.g. if I call `fc.array(fc.integer(10, 20).unique(), 10, 10).sample(2)`, is the `fc.array` implementation allowed to take a `.sample(20)` out of the element arbitrary?).

I'd add distinctness as an argument for `fc.array` or, if we want a nicer API, as a combinator specific to array arbitraries.

arestivo:
> (e.g. if I call fc.array(fc.integer(10, 20).unique(), 10, 10).sample(2), is the fc.array implementation allowed to take a .sample(20) out of the element arbitrary?).

What happens currently is that the `ArbitraryArray` takes two `sample(10)` from the `UniqueArbitrary(ArbitraryInteger)` resulting in two arrays of unique integers.

ruippeixotog:
> What currently happens is that the `ArbitraryArray` takes two `sample(10)` from the `UniqueArbitrary(ArbitraryInteger)` resulting in two arrays of unique integers.

That's an implementation that works well with the example, but there's nothing that prevents us (formal API or informal contract for `sample`) from implementing it later as `.sample(20)` for efficiency purposes. We shouldn't assume a model is sound based on the implementation details today and people shouldn't have to read the implementation to see how a function can and cannot be used.

I feel the API for `Arbitrary` is missing some strong wording on what each function is expected to do; I started adding some descriptions in #108, but there use cases that are still unclear (like this one).

I understand that we may want to take samples with or without replacement internally, but my preference is for every arbitrary to have a way to sample with and without replacement (different methods?). A wrapper that changes the behavior of `sample` seems much trickier.

arestivo:
By the way, just noticed we already have an _ArbitrarySet_.

RuiGuedes:
@ruippeixotog I understand your point of view and I don't know at which point the feature I have been working on, on PR #117, does not fulfill what you mean. By being able to select and configure multiple strategies, we can select a strategy that uses or not ArbitraryUnique.

arestivo:
@RuiGuedes this has nothing to do with strategies. I might just want an array of unique strings, for example.

But I'm starting to agree with @ruippeixotog. We are currently having some problems with the code I wrote [here](https://github.com/fluent-check/fluent-check/pull/120#issuecomment-792265751) that are exactly a reflection of what he was arguing. 

If we have:

```
const a = fc.array(fc.integer(10, 20).unique()
``` 
Then `a.cornerCases()` and `a.canGenerate(v)` are very tricky to get right.

PR #126, opened 2021-03-07
Merge duplicated code used in forall and exists.

hugoferreira:
A lot of code in `FluentCheckUniversal` and `FluentCheckExistential` was basically the same. Abstracted that away in `FluentCheckQuantifier`.

### DIFF
--- a/src/FluentCheck.ts
+++ b/src/FluentCheck.ts
@@ -164,7 +164,7 @@ abstract class FluentCheckQuantifier<K extends string, A, Rec extends ParentRec 
     partial: FluentResult | undefined = undefined,
     depth = 0): FluentResult {
 
-    const example = partial || new FluentResult(this.defaultValue)
+    const example = partial || new FluentResult(!this.breakCondition)
 
     const collection = depth === 0 ?
       this.cache :
@@ -184,21 +184,16 @@ abstract class FluentCheckQuantifier<K extends string, A, Rec extends ParentRec 
     return example
   }
 
-  abstract defaultValue: boolean
   abstract breakCondition: boolean
 }
 
 class FluentCheckUniversal<K extends string, A, Rec extends ParentRec & Record<K, A>, ParentRec extends {}>
   extends FluentCheckQuantifier<K, A, Rec, ParentRec> {
-
-  defaultValue = true
   breakCondition = false
 }
 
 class FluentCheckExistential<K extends string, A, Rec extends ParentRec & Record<K, A>, ParentRec extends {}>
   extends FluentCheckQuantifier<K, A, Rec, ParentRec> {
-
-  defaultValue = false
   breakCondition = true
 }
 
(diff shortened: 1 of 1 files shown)


arestivo:
Maybe use _s_ instead of _b_ as the parameter for breakCondition (as in _satisfiable_).

And maybe _check_ instead of _breakCondition_. 

Rationale: we are checking for satisfiability or non-satisfiability.

hugoferreira:
Would that read well at the call site? We have:

```
if (this.breakCondition(result.satisfiable)) {
 return ...
}
```

Which makes sense... it is the condition that _stops_ the check, not the condition that _continues_  the step. Would you prefer a double negation, like, having this at the call site:

```
if (!this.check(result.satisfiable)) {
 return ...
}
```

... and changing the `forall` condition to:

```
check = (s => !s)
```

... it seems odd, right? We are checking for non-satisfiability in `forall`? And for satisfiability in `exists`? I really think that what we are doing is really early stopping.

RuiGuedes:
I understand your points but to be honest it makes more sense to me to use @hugoferreira proposal because indeed we are really trying to early stop by shrinking. Also since there is already a `check()` function using check could lead to misinterpretations.

src/FluentCheck.ts
```diff
@@ -175,58 +175,31 @@ class FluentCheckUniversal<K extends string, A, Rec extends ParentRec & Record<K
...
-    this.cache = this.dedup.sampleWithBias(this.configuration.sampleSize)
-  }
-
-  protected run(
-    testCase: FluentPicks,
-    callback: (arg: FluentPicks) => FluentResult,
-    partial: FluentResult | undefined = undefined,
-    depth = 0): FluentResult {
+  abstract defaultValue: boolean
+  abstract breakCondition: (b: boolean) => boolean
```
ruippeixotog, line 188: It looks simpler if we just have a `breakValue: boolean` here and do `result.satisfiable === breakValue`.

hugoferreira:
Done. The only redundant thing I now see is that the initial condition is the negation of the break condition. You guys wanna leave this as it is, or make that unification?

hugoferreira:
I also deleted the `initialValue` as it can be correctly derived from the break condition.

RuiGuedes (approved):

src/FluentCheck.ts
```diff
@@ -175,58 +175,26 @@ class FluentCheckUniversal<K extends string, A, Rec extends ParentRec & Record<K
...
-      testCase[this.name] = tp
-      const result = callback(testCase)
-      if (result.satisfiable) {
-        result.addExample(this.name, tp)
-        return this.run(testCase, callback, result, depth + 1)
-      }
-    }
+class FluentCheckUniversal<K extends string, A, Rec extends ParentRec & Record<K, A>, ParentRec extends {}>
+  extends FluentCheckQuantifier<K, A, Rec, ParentRec> {
+  breakCondition = false
```
ruippeixotog, line 192: Now that it's a boolean I prefer if it is called `breakValue` as suggested rather than `breakCondition`.

ruippeixotog (changes requested):

arestivo (approved):
Still not completely happy with the `breakValue` thingy, but I'll allow it.

PR #127, opened 2021-03-07
Simplified Shrink in ArbitraryTuple

hugoferreira:
This is just another step in simplifying the shrink logic of `ArbitraryTuple` after 5467ce8. 

By the way, is anyone brave enough to put some sanity in the types of this class? Typescript [supports variadic tuple types](https://devblogs.microsoft.com/typescript/announcing-typescript-4-0/#variadic-tuple-types) ever since v4.0, but we are clearly not using them correctly, and so we have a lot of unsafe casting going around.


arestivo:
We should have an issue for fixing the types in this class. Otherwise, LGTM.

arestivo (approved):

RuiGuedes (approved):
LGTM

PR #131, opened 2021-03-07
Removed Unique Arbitrary

arestivo:
Removed unique arbitrary in favor of sample methods that are able to sample without replacement in the Arbitrary class.

This means there are 4 different methods for sampling:
- sample()
- sampleWithBias()
- sampleUnique() (was _sampleWithoutReplacement_ originally)
- sampleUniqueWithBias() (was _sampleWithoutReplacementWithBias_ originally)

Maybe this should be changed to a single sample method with parameters for bias and replacement!

FluentCheck has also been changed so that universal and existential checks do not store a unique arbitrary anymore. Instead, they call the sampleUniqueWithBias() method of the original arbitrary.

Tests have been changed to accommodate these changes.


test/arbitrary.test.ts
```diff
@@ -286,15 +280,21 @@ describe('Arbitrary tests', () => {
     })
 
     it('should return a unique sample with bias even with a small sample', () => {
-      expect(fc.scenario()
+      console.log(fc.scenario()
```
RuiGuedes, line 283: Should be deleted.

RuiGuedes (approved):
LGTM and I agree that maybe the 4 different methods for sampling could be changed to a single sample method with parameters for bias and replacement. However, I think that for now, we leave things as they are and in the future, if needed, we change how things are being done.

test/arbitrary.test.ts
```diff
@@ -417,7 +411,6 @@ describe('Arbitrary tests', () => {
     })
 
     it('should remain no arbitrary when compose with unique, map, and filter', () => {
```
hugoferreira, line 413: Can we change the description here to not have unique?

hugoferreira (changes requested):

src/arbitraries/Arbitrary.ts
```diff
@@ -44,7 +46,34 @@ export abstract class Arbitrary<A> {
   }
 
   /**
-   * The special cases for this arbitrary, which can be used during sampling to give higher weight to certain elements.
+   * Returns a sample of picks of a given size without replacements.
+   *
+   * While this function doesn't pick corner cases, each pick is not necessarily independent
```
ruippeixotog, line 51: This description does not look right given that we actually require all elements to be unique here. Can you update it accordingly?

src/arbitraries/Arbitrary.ts
```diff
@@ -44,7 +46,34 @@ export abstract class Arbitrary<A> {
...
+   * from the others: different arbitraries might choose to use smarter search strategies to
+   * maximize the probability of falsifying a property, e.g. by including only unique elements.
+   */
+  sampleWithoutReplacement(sampleSize = 10, cornerCases: FluentPick<A>[] = []): FluentPick<A>[] {
```
ruippeixotog, line 55: nit: `sampleUnique` is more concise and more user friendly for those without statistical background. I'm ok either way though.

ruippeixotog (changes requested):
LGTM overall. This also makes it easier to integrate my implementations of indexed arbitraries in #28, which should provide efficient `sampleWithoutReplacement` implementations for most primitive and collection types.

hugoferreira (approved):
LGTM

hugoferreira:
Bringing back #28 would be amazing...

arestivo:
> This also makes it easier to integrate my implementations of indexed arbitraries ...

I knew you had ulterior motives!

src/arbitraries/Arbitrary.ts
arestivo: @hugoferreira Should we use _unique_ or _withoutReplacement_?

hugoferreira: I'm ok with `sample` and `sampleUnique` 👍

hugoferreira (approved):

ruippeixotog (approved):

src/arbitraries/UniqueArbitrary.ts
```diff
@@ -1,48 +0,0 @@
...
-import {Arbitrary, WrappedArbitrary} from './internal'
-import {stringify} from './util'
-
-export class UniqueArbitrary<A> extends WrappedArbitrary<A> {
-  constructor(readonly baseArbitrary: NonNullable<Arbitrary<A>>) {
-    super(baseArbitrary)
-  }
-
-  sample(sampleSize = 10, cornerCases: FluentPick<A>[] = []): FluentPick<A>[] {
-    // TODO: Here lies dragons! If you see start seeing things in double when
```
ruippeixotog, line 11: @arestivo I noticed now you removed this comment. I think it's important to keep this caveat visible as unique only works if values are deemed equal under our own `stringify`, which uses custom logic.

PR #141, opened 2021-03-08
Control seed of RNG and output it on result #139

### LINKED ISSUE #139: Control seed of RNG and output it on result
Debugging non-deterministic checks is becoming hard, and we sometimes want to replicate a specific case generation. I suggest the following:

1. Allow the user to provide the initial seed for the internal RNG via scenario configuration
2. Return the used seed in the result, so that we can replicate non-deterministic fails, such as the one in #138

tiagoacastro:
WIP closes #139


src/arbitraries/util.ts
```diff
@@ -13,6 +13,14 @@ export function stringify(object: any) {
   return (object instanceof Object || object instanceof Array) ? JSON.stringify(object) : object
 }
 
+//Splitmix64 code adapted from https://prng.di.unimi.it/splitmix64.c
```
arestivo, line 16: Why are we not using a library implementing splitmix64? For example https://www.npmjs.com/package/@aureooms/js-pseudo-random

The code is always the same, but still...
https://aureooms.github.io/js-pseudo-random/file/src/splitmix64.js.html

This is an honest question, do we want a new dependency because of 5 lines of code?

arestivo: Also, why are we using splitmix64? I only talked about this particular RNG after looking at some nice pictures. You should investigate others...

hugoferreira: Actually, you didn't even mentioned splitmix64, but xoshiro128.

tiagoacastro: The creator of xoshiro mentions the use of splitmix64 for the creation of the seed, but I believe that is because Splitmix64 is the default pseudo-random number generator algorithm in Java and the code on the creator's website is in java, so I removed the use of this particular algorithm. As to the choice of xoshiro, after browsing a bit, according to a comment on [this thread](https://stackoverflow.com/questions/521295/seeding-the-random-number-generator-in-javascript), It is the fastest generator that offers a 128-bit state.

src/arbitraries/Arbitrary.ts
```diff
@@ -115,4 +116,6 @@ export abstract class Arbitrary<A> {
   chain<B>(f: (a: A) => Arbitrary<B>): Arbitrary<B> { return new ChainedArbitrary(this, f) }
 
   toString(depth = 0): string { return ' '.repeat(depth * 2) + `Base Arbitrary: ${this.constructor.name}`  }
+
+  setSeed(seed: bigint): void { this.seed = seed }
```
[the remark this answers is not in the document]
tiagoacastro: The setseed function serves the purpose of passing the scenario seed to the arbitraries.  The change you are proposing is relative to the way the user specifies that he wants a certain seed to be used. However, I understand what you mean and will make the change.

tiagoacastro:
> Why are we receiving a string and then converting it to a number? Why not receive the number right away?

I don't understand to which part of the code you are referring, the user specifies the seed in hexadecimal via string, which is then converted to hexadecimal, stored, and used to create the generator. The reason why it is specified in hexadecimal is that the seeds being generated have 128 bits.

> Why do we need the getter?

I assume you are referring to the getParent function? it is so the generator is propagated upwards (and not only to the direct parent) in case the user specifies withseed after the arbitraries.

> I reallly dislike having a null generator, and decide to use one internally. Why not make the generator a `() => number` and initialize it to `Math.random()`. Then we wouldn't need all this duplicated code.

I don't understand how that would work since Math.random() straight up generates a value no a generator.

> If we only propagate the generator to the parent, then `.withSeed` would need to be the last thing to call?

If it is only propagated to the parent and not iteratively propagated as mentioned before with the use of getParent, then withSeed would have to be specified right after the scenario.

hugoferreira:
> > Why are we receiving a string and then converting it to a number? Why not receive the number right away?
> 
> I don't understand to which part of the code you are referring, the user specifies the seed in hexadecimal via string, which is then converted to hexadecimal, stored, and used to create the generator. The reason why it is specified in hexadecimal is that the seeds being generated have 128 bits.

And why do we need 128 bits of seed? Why not BigInt? I just want to understand the rationale.

> > Why do we need the getter?
> 
> I assume you are referring to the getParent function? it is so the generator is propagated upwards (and not only to the direct parent) in case the user specifies withseed after the arbitraries.

Why are we not accessing the field directly, if all the getter does is expose the field for subclasses?

> > I reallly dislike having a null generator, and decide to use one internally. Why not make the generator a `() => number` and initialize it to `Math.random()`. Then we wouldn't need all this duplicated code.
> 
> I don't understand how that would work since Math.random() straight up generates a value no a generator.

Math.random has type `() => number`. That all we need: a function that receives zero parameters and returns a number. We don't specifically need a `generator` to generate random numbers at the call sites. 
 

> > If we only propagate the generator to the parent, then `.withSeed` would need to be the last thing to call?
> 
> If it is only propagated to the parent and not iteratively propagated as mentioned before with the use of getParent, then withSeed would have to be specified right after the scenario.

Ok, show that to me with a test and I'll be convinced :)

test/generation.test.ts
```diff
@@ -0,0 +1,45 @@
...
+
+  it('Generator propagates to arbitraries with seed specification', () => {
+    const arb1: Arbitrary<number> = fc.integer(-10, 10)
+    const arb2: Arbitrary<number> = fc.integer(-10, 10)
+    const fs = fc.scenario().forall('a', arb1).withSeed('aaa').forall('b', arb2)
+
+    expect(fs.generator === arb1.generator).to.be.true
+    expect(fs.generator === arb2.generator).to.be.true
+  })
+})
```
hugoferreira, line 45: All this tests test against a specific implementation; they provide evidence that the generator have the same reference... But more usefull is to test against a behavior: to prove that the outcome is what we are expecting. Think about it: when you create a `fc.scenario()` and set the seed, _what you want_ is that different runs of the same scenario produce the same samples for the same set of arbitraries. Can you prove that with these tests?

hugoferreira (changes requested):
Improve these tests so that they are remain behaviourally valid and not implementation-specific.

tiagoacastro:
> > > Why are we receiving a string and then converting it to a number? Why not receive the number right away?
> > 
> > 
> > I don't understand to which part of the code you are referring, the user specifies the seed in hexadecimal via string, which is then converted to hexadecimal, stored, and used to create the generator. The reason why it is specified in hexadecimal is that the seeds being generated have 128 bits.
> 
> And why do we need 128 bits of seed? Why not BigInt? I just want to understand the rationale.
> 
> > > Why do we need the getter?
> > 
> > 
> > I assume you are referring to the getParent function? it is so the generator is propagated upwards (and not only to the direct parent) in case the user specifies withseed after the arbitraries.
> 
> Why are we not accessing the field directly, if all the getter does is expose the field for subclasses?
> 
> > > I reallly dislike having a null generator, and decide to use one internally. Why not make the generator a `() => number` and initialize it to `Math.Random()`. Then we wouldn't need all this duplicated code.
> > 
> > 
> > I don't understand how that would work since Math.random() straight up generates a value no a generator.
> 
> `Math.Random()` has type `() => number`. That's all we need: a function that receives zero parameters and returns a number. We don't specifically need a `generator` to generate random numbers at the call sites. I would argue for the same thing @ruippeixotog argued in #139. Build a means for the user to provide a `(seed: number) => () => number` generator as an alternative, allowing them to weigh the tradeoffs themselves. This way, you don't even need to worry about using `splitmix64`, `xoshiro128` or any other library; that will be up to the user. More specifically, we are arguing that the following signature could be used:
> 
> ```
> withSeed(generator: (seed: number) => () => number = Math.Random): FluentCheckSeed<Rec, ParentRec>
> ```
> 
> and that your `.setGenerator()` (or get rid of this and receive the RNG via the constructor with a default) in `Arbitrary` could become:
> 
> ```
> setGenerator(generator: () => number): void { this.generator = generator }
> ```
> 
> which would make your particular `pick()` in `ArbitraryInteger`:
> 
> ```
> pick() {
>   const value = Math.floor(this.generator()) * (this.max - this.min + 1)) + this.min
>   return {value, original: value}
> }
> ```
> 
> And if you do that at the `Arbitrary` base level, then you won't even need to do it for the others. Just make sure that instead of calling `Math.random()`, they are calling the `this.generator` (or a better name).
> 
> > > If we only propagate the generator to the parent, then `.withSeed` would need to be the last thing to call?
> > 
> > 
> > If it is only propagated to the parent and not iteratively propagated as mentioned before with the use of getParent, then withSeed would have to be specified right after the scenario.
> 
> Ok, show that to me with a test and I'll be convinced :)

1. I understood incorrectly what the creator of xoshiro mentions on its website, he talks about 64 bit seeds so I have changed to that. As to why the seed is stored as a string and not a bigint is because the seed will never be needed in bigint form, only in string form because that is how the package used for xoshiro128** receives the strings (https://www.npmjs.com/package/rand-seed).
2. You are right, it doesn't make sense.

src/FluentCheck.ts
```diff
@@ -13,11 +13,20 @@ class FluentResult {
   }
 }
 
-export type FluentConfig = { sampleSize?: number, shrinkSize?: number }
+export type FluentConfig = { sampleSize?: number, shrinkSize?: number}
 
 export class FluentCheck<Rec extends ParentRec, ParentRec extends {}> {
-  constructor(protected readonly parent: FluentCheck<ParentRec, any> | undefined = undefined,
+  public prng: {
```
arestivo, line 19: You should define a type (using a TS type alias) instead of repeating this over and over!

test/generation.test.ts
```diff
@@ -0,0 +1,100 @@
...
+    const arb2: Arbitrary<number>= fc.integer(-10, 10)
+    fc.scenario().forall('a', arb1).withGenerator(prng, 1234).forall('b', arb2)
+    const sample1: FluentPick<number>[] = arb1.sampleUniqueWithBias()
+    const sample2: FluentPick<number>[] = arb2.sampleUniqueWithBias()
+
+    const arb3: Arbitrary<number>= fc.integer(-10, 10)
+    const arb4: Arbitrary<number>= fc.integer(-10, 10)
+    fc.scenario().forall('a', arb3).withGenerator(prng, 1234).forall('b', arb4)
+    const sample3: FluentPick<number>[] = arb3.sampleUniqueWithBias()
+    const sample4: FluentPick<number>[] = arb4.sampleUniqueWithBias()
```
hugoferreira, line 95: Is there a reason you don't trust typescript type inference? xD

src/FluentCheck.ts
```diff
@@ -13,11 +13,20 @@ class FluentResult {
...
+    generator: () => number,
+    seed: number | undefined
+  }
+
+  constructor(readonly parent: FluentCheck<ParentRec, any> | undefined = undefined,
     public readonly configuration: FluentConfig = {sampleSize: 1000, shrinkSize: 500}) {
+    if (this.parent === undefined)
```
arestivo, line 26: Use null coalescing to simplify this:

https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#nullish-coalescing

src/FluentCheck.ts
```diff
@@ -231,3 +252,24 @@ class FluentCheckAssert<Rec extends ParentRec, ParentRec extends {}> extends Flu
...
+
+class FluentCheckGenerator<Rec extends ParentRec, ParentRec extends {}> extends FluentCheck<Rec, ParentRec> {
+  constructor(
+    readonly parent: FluentCheck<ParentRec, any>,
+    config: FluentConfig,
+    readonly generator: (seed: number) => () => number,
+    readonly seed?: number
+  ) {
+    super(parent, config)
+    if (seed !== undefined)
```
arestivo, line 264: Use nullish coalescing

src/FluentCheck.ts
```diff
@@ -231,3 +252,24 @@ class FluentCheckAssert<Rec extends ParentRec, ParentRec extends {}> extends Flu
...
+      this.prng.seed = seed
+    else
+      this.prng.seed = Math.floor(Math.random() * 0x100000000)
+    this.prng.generator = generator(this.prng.seed)
```
arestivo, line 268: This should be instantiated only when the scenario is run so that two runs of the same seeded scenario give the same result.

src/arbitraries/Arbitrary.ts
```diff
@@ -3,6 +3,7 @@ import {ChainedArbitrary, FilteredArbitrary, MappedArbitrary, NoArbitrary} from
 import {stringify} from './util'
 
 export abstract class Arbitrary<A> {
+  generator: () => number = Math.random
```
arestivo, line 6: The generator should be passed to the sample and sampleWithBias functions instead of being part of an arbitrary.

hugoferreira (changes requested):
I'm not sure we don't have the following problem (pseudo-code):

```
const s1 = fc.scenario().withGenerator(..., 1234).given('a', fc.integer())
s1.run().tracesOf('a') === s1.run().tracesOf('a')
```

In other words, running the same scenario twice should produce the same result. Because the way you call the generator factory in the setter (instead of instiantiating it in the `run/check`) I believe this would be false.

src/FluentCheck.ts
hugoferreira: I've provided the same feedback in my review. I think this can be shown with a test, although we are surely missing a way to trace all the generated `testCases`.

package.json
```diff
@@ -17,6 +17,8 @@
   "author": "André Restivo, Hugo Sereno Ferreira, Rui Gonçalves",
   "license": "MIT",
   "dependencies": {
+    "rand-seed": "^1.0.0",
+    "set-blocking": "^2.0.0",
```
hugoferreira, line 21: So we need these in `dependencies`? If they are now used only for testing, can't we move them to `devDependencies`

hugoferreira (changes requested):

package.json
tiagoacastro: No, I forgot to remove them because I didn't use them after all, thank you for pointing it out

src/FluentCheck.ts
tiagoacastro: I didn't know this existed but it's interesting and I will start using it in the future, however, I don't believe this a case where it can be used because this.parent is the one being checked if it is set and this.prng is the one being assigned. Maybe I understood incorrectly how nullish coalescing works?

src/FluentCheck.ts
tiagoacastro: Can it be used here as well? the variable being checked is the seed passed by parameter and the one being assigned is the seed present in the prng variable

src/FluentCheck.ts
arestivo: Not sure if this works:

`this.parent?.prng || { generator: Math.random, seed: undefined }`

arestivo: Also, take a look at this for future reference:

https://linguinecode.com/post/how-to-solve-typescript-possibly-undefined-value

src/arbitraries/Arbitrary.ts
```diff
@@ -15,7 +13,7 @@ export abstract class Arbitrary<A> {
   /**
    * Generates a random element. This operation is stateless.
    */
-  abstract pick(): FluentPick<A> | undefined
+  abstract pick(generator: () => number): FluentPick<A> | undefined
```
hugoferreira, line 16: Please provide a default generator...

src/arbitraries/Arbitrary.ts
```diff
@@ -31,11 +29,11 @@ export abstract class Arbitrary<A> {
    * Returns a sample of picks of a given size. Sample might contain repeated values
    * and corner cases are not taken into account.
    */
-  sample(sampleSize = 10): FluentPick<A>[] {
+  sample(sampleSize = 10, generator: () => number): FluentPick<A>[] {
```
hugoferreira, line 32: Please provide a default generator...

src/arbitraries/Arbitrary.ts
```diff
@@ -47,7 +45,7 @@ export abstract class Arbitrary<A> {
    * Returns a sample of picks of a given size withour replacement. Sample will
    * not contain repeated values. Corner cases are not taken into account.
    */
-  sampleUnique(sampleSize = 10, cornerCases: FluentPick<A>[] = []): FluentPick<A>[] {
+  sampleUnique(sampleSize = 10, generator: () => number, cornerCases: FluentPick<A>[] = []): FluentPick<A>[] {
```
hugoferreira, line 48: Please provide a default generator _and_ probably make it the last argument... arguments _with defaults_ should be ordered by likelihood of having an argument passed, and in case of a tie, in the end so as to avoid changing unnecessary code.

test/arbitrary.test.ts
```diff
@@ -406,8 +406,8 @@ describe('Arbitrary tests', () => {
     })
 
     it('should return an empty sample', () => {
-      expect(fc.empty().sample().length).to.eq(0)
-      expect(fc.empty().sampleWithBias().length).to.eq(0)
+      expect(fc.empty().sample(10, Math.random).length).to.eq(0)
+      expect(fc.empty().sampleWithBias(10, Math.random).length).to.eq(0)
     })
 
     it('should remain no arbitrary when compose with map, and filter', () => {
```
hugoferreira, line 413: Revert all changes here

test/generation.test.ts
```diff
@@ -26,72 +26,40 @@ describe('Generation tests', () => {
...
-
   it('Generator generates different values for two similar arbitraries without generator specification', () => {
-    const arb1 = fc.integer(-10, 10)
-    const arb2 = fc.integer(-10, 10)
-    fc.scenario().forall('a', arb1).forall('b', arb2)
+    const sc1 = fc.scenario().forall('a', fc.integer(-10, 10))
+    const sc2 = sc1.forall('b', fc.integer(-10, 10))
 
-    expect(arb1.sampleUniqueWithBias() !== arb2.sampleUniqueWithBias()).to.be.true
+    expect(sc1.getCache() !== sc2.getCache()).to.be.true
```
hugoferreira, line 33: What tyoe of object does `getCache()` return? Are you sure `!==` is what you want (i.e. to compare references and not content?)

test/generation.test.ts
```diff
@@ -26,72 +26,40 @@ describe('Generation tests', () => {
...
 
   it('Generator generates different values for two similar arbitraries with generator specification', () => {
-    const arb1 = fc.integer(-10, 10)
-    const arb2 = fc.integer(-10, 10)
-    fc.scenario().forall('a', arb1).withGenerator(prng).forall('b', arb2)
+    const sc1 = fc.scenario().withGenerator(prng).forall('a', fc.integer(-10, 10))
+    const sc2 = sc1.forall('b', fc.integer(-10, 10))
 
-    expect(arb1.sampleUniqueWithBias() !== arb2.sampleUniqueWithBias()).to.be.true
+    expect(sc1.getCache() !== sc2.getCache()).to.be.true
```
hugoferreira, line 52: Same as above...

test/generation.test.ts
```diff
@@ -26,72 +26,40 @@ describe('Generation tests', () => {
...
-    fc.scenario().forall('a', arb3).withGenerator(prng, 1234).forall('b', arb4)
-    const sample3 = arb3.sampleUniqueWithBias()
-    const sample4 = arb4.sampleUniqueWithBias()
+    const sc3 = fc.scenario().withGenerator(prng, 1234).forall('a', fc.integer(-10, 10))
+    const sc4 = sc3.forall('b', fc.integer(-10, 10))
 
-    expect(compareSamples(sample1, sample3)).to.be.true
-    expect(compareSamples(sample2, sample4)).to.be.true
+    expect(compareSamples(sc1.getCache(), sc3.getCache())).to.be.true
+    expect(compareSamples(sc2.getCache(), sc4.getCache())).to.be.true
```
hugoferreira, line 63: Same as above...

hugoferreira (changes requested):

hugoferreira:
Please note this branch doesn't seem to be in sync with master. You should probably rebase it.

test/generation.test.ts
```diff
@@ -0,0 +1,65 @@
...
+  }
+}
+
+function compareSamples(sample1: FluentPick<number>[], sample2: FluentPick<number>[]): boolean {
+  for (let i = 0; i < sample1.length; i++) {
+    if (sample1[i].value !== sample2[i].value || sample1[i].original !== sample2[i].original)
+      return false
+  }
+  return true
+}
```
hugoferreira, line 18: I'm sure `mocha` has functions to make deep comparisons of arrays of objects. Use it.

hugoferreira (dismissed):

test/arbitrary.test.ts
```diff
@@ -291,7 +291,7 @@ describe('Arbitrary tests', () => {
   describe('Chained Arbitraries', () => {
     it('should allow the creation of array with size based on an integer arbitrary', () => {
       expect(
-        fc.integer(2, 2).chain(i => fc.array(fc.constant(i), i, i)).pick()!.value
+        fc.integer(2, 2).chain(i => fc.array(fc.constant(i), i, i)).pick(Math.random)!.value
```
arestivo, line 294: Why are we changing the RNG in this particular test?

src/strategies/FluentStrategyMixins.ts
```diff
@@ -49,7 +49,8 @@ export function Cached<TBase extends MixinStrategy>(Base: TBase) {
 export function Biased<TBase extends MixinStrategy>(Base: TBase) {
   return class extends Base {
     buildArbitraryCollection<A>(arbitrary: Arbitrary<A>, sampleSize = this.configuration.sampleSize): FluentPick<A>[] {
-      return this.isDedupable() ? arbitrary.sampleUniqueWithBias(sampleSize) : arbitrary.sampleWithBias(sampleSize)
+      return this.isDedupable() ? arbitrary.sampleUniqueWithBias(sampleSize, this.prng.generator) :
```
arestivo, line 52: Maybe a newline after "?"

test/arbitrary.test.ts
tiagoacastro: Because fc.array returns an object of type Arbitrary and since the pick method in Arbitrary is marked as abstract it cannot have a default value for its parameters. A solution is change the code for the pick method in the Arbitrary class to:

pick(generator: () => number = Math.random): FluentPick<A> | undefined { return }

Which version do you prefer?

src/strategies/FluentStrategyMixins.ts
tiagoacastro: The second line is still too long according to lint standards, that's why I did it like this, I can however turn it into 3 different lines one for the check, one for the true case, and another for the false case.

hugoferreira (approved):
We've refactored the code so that this issue gets merged asap. @ruippeixotog , @RuiGuedes , @tiagoacastro please review.

RuiGuedes (approved):
LGTM!

PR #146, opened 2021-03-11
Code constants extraction-based search strategy implementation

### LINKED ISSUE #130: Code constants extraction-based search strategy implementation
After #117 being merged the first informed strategy to be implemented is based on extracting constants from code, either a file or function and use such constant values as inputs values for testing.

RuiGuedes:
Closes #130. Comments and explanations in the code.


hugoferreira (changes requested):
...

src/strategies/mixins/FluentConstantExtractionMixin.ts
```diff
@@ -36,11 +36,18 @@ export function ConstantExtractionBased<TBase extends MixinStrategy>(Base: TBase
       })
 
       const numericsAndPunctuators = filteredTokens.reduce(function (acc, token, index) {
-        if (token.type === 'Numeric')
+        if (token.type === 'Numeric') {
+          const leftPunctuatorIndex = (filteredTokens[index - 1] !== undefined &&
+            filteredTokens[index - 1].value !== '-') ? 1 : 2
```
hugoferreira, line 41: I'm starting to feel that a token-based parser would not scale very well...

RuiGuedes: I understand your point but apart from `numeric` we are only interested in `string` tokens. So I think we are good.

RuiGuedes: Btw this code section you used for making your statement has only two types of tokens: `punctuator` or `numeric`

RuiGuedes:
Depends on #153 and #156

src/strategies/FluentStrategy.ts
```diff
@@ -15,17 +15,28 @@ export class FluentStrategy implements FluentStrategyInterface {
    */
   public arbitraries: StrategyArbitraries = {}
 
+  /**
+   * Contains all the assertions concercing a test case
```
arestivo, line 19: concerning

src/strategies/FluentStrategy.ts
```diff
@@ -53,34 +66,45 @@ export class FluentStrategy implements FluentStrategyInterface {
...
   /**
    * Hook that acts as point of extension of the configArbitrary function and that enables an arbitrary to be shrinked.
    */
   shrink<K extends string>(_name: K, _partial: FluentResult | undefined) {}
 
+  /**
+   * Hook that acts as point of extension of the buildArbitraryCollection function and that enables the strategy to use
+   * extracted constants from code in the test cases.
+   */
+  getArbitraryExtractedConstants<A>(_arbitrary: Arbitrary<A>): FluentPick<A>[] {
```
arestivo, line 90: Can this hook be used for other purposes than extracting constants?

test/strategies.test.ts
```diff
@@ -0,0 +1,63 @@
...
+          .withConstantExtraction()
+        )
+        .forall('a', fc.integer())
+        .forall('b', fc.integer())
+        .then(({a, b}) => a + b === 100 ? a + b === b : a + b === b + a)
+        .check()
+      ).to.have.property('satisfiable', false)
+    })
+
+    it('should be able to extract constants and compute numeric range from assertion', () => {
```
arestivo, line 22: I still think that your extraction strategy being able to solve this case is too much magic and might come back to bite us in the ass later on.

arestivo (approved):
I left some comments but LGTM for now.

src/strategies/FluentStrategy.ts
RuiGuedes: Already fixed!

src/strategies/FluentStrategy.ts
RuiGuedes: As of right now, it is only being used for that purpose. I think that in the future and according to our needs we can rename it to something else.

test/strategies.test.ts
RuiGuedes: @arestivo According to your opinion and also with what I have discussed with @hugoferreira at our last meeting I have decided to make this PR as simple as possible and therefore removed the numeric range computation as well as the string mutation. 

As such, the constants extraction is simple as extracting numbers and strings, where for numbers, it extracts the value and creates two other values that are immediately above and below, and for strings, it extracts the value and creates two other strings that are the substrings of the original one and with half the size.

The developed code for string manipulation and numerics range computation (now removed) was not completely deleted because I decided to save it in my gists for future use if needed. 

The idea now (after this PR being merged) is to create a synthetic test suite related to constants extraction and as we identify new necessities we expand/enhance the constant extraction feature.

RuiGuedes:
@hugoferreira @arestivo When I created this PR my intention was to merge it on `msc-ruiguedes-dev`. However, since `master` is always evolving there will be always a lot of conflicts to solve (which can take a lot of time as I experienced). Therefore, can I change the base branch to be the `master` instead of `msc-ruiguedes-dev`, or there is any problem?

PR #151, opened 2021-03-15
Fixing ArbitraryInteger shrinking problem

RuiGuedes:
The current version of the ArbitraryIntegers shrinking process is not working properly because it does not allow shrinking an integer to its neutral value `0`. The integers' test suite version on the `master` branch only succeeds because the current default strategy considers bias and `0` belongs to the set of corner cases.

### DIFF
--- a/src/arbitraries/ArbitraryInteger.ts
+++ b/src/arbitraries/ArbitraryInteger.ts
@@ -28,14 +28,14 @@ export class ArbitraryInteger extends Arbitrary<number> {
       const lower = Math.max(0, this.min)
       const upper = Math.max(lower, initial.value - 1)
 
-      if (lower === upper) return NoArbitrary
+      if (lower === upper && initial.value !== 1) return NoArbitrary
 
       return fc.integer(lower, upper)
     } else if (initial.value < 0) {
       const upper = Math.min(0, this.max)
       const lower = Math.min(upper, initial.value + 1)
 
-      if (lower === upper) return NoArbitrary
+      if (lower === upper && initial.value !== -1) return NoArbitrary
 
       return fc.integer(lower, upper)
     }
--- a/src/arbitraries/ArbitraryInteger.ts
+++ b/src/arbitraries/ArbitraryInteger.ts
@@ -26,16 +26,16 @@ export class ArbitraryInteger extends Arbitrary<number> {
   shrink(initial: FluentPick<number>): Arbitrary<number> {
     if (initial.value > 0) {
       const lower = Math.max(0, this.min)
-      const upper = Math.max(lower, initial.value - 1)
+      const upper = initial.value - 1
 
-      if (lower === upper && initial.value !== 1) return NoArbitrary
+      if (upper < lower) return NoArbitrary
 
       return fc.integer(lower, upper)
     } else if (initial.value < 0) {
       const upper = Math.min(0, this.max)
-      const lower = Math.min(upper, initial.value + 1)
+      const lower = initial.value + 1
 
-      if (lower === upper && initial.value !== -1) return NoArbitrary
+      if (lower > upper) return NoArbitrary
 
       return fc.integer(lower, upper)
     }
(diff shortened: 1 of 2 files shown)


RuiGuedes:
The following test with the following configuration allows you to replicate the stated issue.

image

image

src/arbitraries/ArbitraryInteger.ts
```diff
@@ -28,14 +28,14 @@ export class ArbitraryInteger extends Arbitrary<number> {
       const lower = Math.max(0, this.min)
       const upper = Math.max(lower, initial.value - 1)
 
-      if (lower === upper) return NoArbitrary
+      if (lower === upper && initial.value !== 1) return NoArbitrary
```
hugoferreira, line 31: Are you sure this will work for any integer interval that doesn't overlap 0?

RuiGuedes: Do you mean like [5,10] for example?

RuiGuedes: I confirmed it also works. I used this particular test for making the verification

```
it('finds that summing two positive numbers in a range nevers returns zero', () => {
    expect(fc.scenario()
      .config(fc.strategy()
        .withRandomSampling()
        .usingCache()
        .withoutReplacement()
        .withShrinking()
      )
      .forall('a', fc.integer(5, 10))
      .exists('b', fc.integer(1, 2))
      .then(({a, b}) => a + b === 0)
      .check()
    ).to.have.property('satisfiable', false)
  })
```

I think the modification is safe because it uses directly the `initial.value` and not the `upper` and `lower` limits.

RuiGuedes: Let me know if have any more doubts!

hugoferreira: I am sorry, but I am still not conviced this actually works for all cases 😅 @arestivo, @ruippeixotog could you do a sanity check here? What makes `-1` and `1` special when shrinking an `fc.integer(5, 10)`?

ruippeixotog: There's nothing special about `-1` and `1`, I think this PR just special cases the shrinker for `fc.integer(0, n)` ranges, where it will now shrink to 0. My guess is that if you try to shrink some value in `fc.integer(5, 10)` it will still shrink to 6, not 5 as intended.

Without compiling or testing anything, the real problem seems to be the `lower === upper` part. In those cases we are incorrectly returning `NoArbitrary` (which means that we didn't find any "smaller" input). The condition should be changed to `lower > upper` and `upper` should be defined simply as `initial.value - 1`.

RuiGuedes: `-1` and `1` were only being _special_ due to the way code was developed, which did not allow to shrink from either `-1` or `1` to `0`.

However, I completely agree with @ruippeixotog's last comment since the _suggested_ modifications make total sense. I tried them and they work. Already pushed a new version of fixing the `ArbitraryInteger` shrinking problem. @hugoferreira  Let me know how you feel about the new solution now.

hugoferreira (approved):
Not it makes sense to me 😄

test/integers.test.ts
```diff
@@ -57,6 +57,12 @@ describe('Integer tests', () => {
 
   it('finds that adding 1000 makes any number larger and shrinks the example', () => {
     expect(fc.scenario()
+      .config(fc.strategy()
```
ruippeixotog, line 60: Any reason why you added this here? If not, let's keep the PR clean and focused on a single thing.

ruippeixotog (approved):
LGTM too :+1: Btw, this bug was caused (or at least more difficult to spot) by not having clear documented expectations for each API method. Just by looking at the signature, it it not clear what `shrink` needs to do - should the returned arbitrary contain the picked value itself, or only an arbitrary of strictly "smaller" elements? For each arbitrary implementation we have, what does it mean to be "smaller"?

This is the kind of things that should be clear in docs. Not having JSDoc or something equivalent on core APIs tinkles with my OCD :)

test/integers.test.ts
RuiGuedes: Yes to make sure the `biased` values are not being considered. This way we ensure that the shrinking process is working as intended.

hugoferreira:
Can we merge this?

PR #153, opened 2021-03-15
Fix canGenerate in MappedArbitrary

### LINKED ISSUE #152: Fix canGenerate in MappedArbitrary
The current mapped arbitrary uses this code to check if a `pick` can be generated:

```
canGenerate(pick: FluentPick<B>) {
    /* && pick.value === this.f(pick.original) */
    return this.baseArbitrary.canGenerate({value: pick.original, original: pick.original})
  }
```

Well, that's not strictly true in cases such as: 

```
expect(fc.integer(0, 1).map(a => a === 1 ? 'true' : 'false')
      .canGenerate({original: 'true', value: 'true'})).to.be.true
```

Or even more complicated:

```
expect(fc.integer(-10, 0).map(a => Math.abs(a))
      .canGenerate({original: -5, value: -5})).to.be.true // Should be false
```

RuiGuedes:
Closes #152


src/arbitraries/MappedArbitrary.ts
```diff
@@ -32,8 +39,17 @@ export class MappedArbitrary<A, B> extends Arbitrary<B> {
...
+      (Array.isArray(this.config.inverseFunction(pick.value)) ?
+        this.config.inverseFunction(pick.value) as A[] :
+        [this.config.inverseFunction(pick.value)])
+
+    for (const value of inverseValues) {
+      if (this.baseArbitrary.canGenerate({value, original: pick.original}))
+        return true
+    }
+
+    return false
```
hugoferreira, line 52: `return inverseValues.some(value => this.baseArbitrary.canGenerate({value, original: pick.original}))` ?

src/arbitraries/MappedArbitrary.ts
```diff
@@ -32,8 +39,17 @@ export class MappedArbitrary<A, B> extends Arbitrary<B> {
   }
 
   canGenerate(pick: FluentPick<B>) {
-    /* && pick.value === this.f(pick.original) */
-    return this.baseArbitrary.canGenerate({value: pick.original, original: pick.original})
+    const inverseValues = !(this.config && this.config.inverseFunction) ? [pick.original] :
+      (Array.isArray(this.config.inverseFunction(pick.value)) ?
+        this.config.inverseFunction(pick.value) as A[] :
+        [this.config.inverseFunction(pick.value)])
```
hugoferreira, line 45: Would this work for arrays generated by an `ArbitraryArray`?

src/arbitraries/types.ts
```diff
@@ -8,3 +8,8 @@ export type ArbitrarySize = {
   type: 'exact' | 'estimated'
   credibleInterval?: [number, number]
 }
+
+export type MappedArbitraryConfig<A,B> = {
+  inverseFunction?: (b: B) => A | A[],
+  customCanGenerate?: (pick: FluentPick<B>) => boolean
```
hugoferreira, line 14: I would simply change the name to `inverseMap` and `canGenerate`. I also don't like the name `MappedArbitraryConfig`, but can't think of a better alternative xD

hugoferreira (changes requested):
My only problem is with `ArbitraryArrays` and inverse functions; a test would convince me that everything is working. Otherwise, it's just small nits and LGTM.

src/arbitraries/types.ts
hugoferreira: By the way, this solution you've conjured is very ingenious, as it supports passing inline overrides in those cases! The downside is that it becomes too verbose...

src/arbitraries/MappedArbitrary.ts
```diff
@@ -32,8 +39,17 @@ export class MappedArbitrary<A, B> extends Arbitrary<B> {
   }
 
   canGenerate(pick: FluentPick<B>) {
-    /* && pick.value === this.f(pick.original) */
-    return this.baseArbitrary.canGenerate({value: pick.original, original: pick.original})
+    const inverseValues = !(this.config && this.config.inverseFunction) ? [pick.original] :
```
arestivo, line 42: I think you can simplify `this.config && this.config.inverseFunction` to `this.config?.inverseFunction`.

src/arbitraries/MappedArbitrary.ts
```diff
@@ -32,8 +39,17 @@ export class MappedArbitrary<A, B> extends Arbitrary<B> {
...
+      (Array.isArray(this.config.inverseFunction(pick.value)) ?
```
arestivo, line 43: Why are you calculating the `inverseFunction` twice?

src/arbitraries/MappedArbitrary.ts
arestivo: I also have this doubt. Why do you need to have these two cases?

test/arbitrary.test.ts
```diff
@@ -236,16 +236,34 @@ describe('Arbitrary tests', () => {
...
+      expect(fc.integer(0, 1)
+        .map(a => a === 1, {inverseFunction: b => b ? 1 : 0})
+        .filter(a => a === false)
+        .map(a => a ? 0 : 1, {inverseFunction: b => b === 0 ? true : false})
         .canGenerate({original: 1, value: 1})).to.be.true
     })
   })
 
+  describe('Mapped Arbitraries', () => {
+    it('should be able to inverse map transformation', () => {
```
arestivo, line 254: Capitalize 'Should'

test/arbitrary.test.ts
```diff
@@ -236,16 +236,34 @@ describe('Arbitrary tests', () => {
 
   describe('Filtered Arbitraries', () => {
     it('A filtered mapped filtered arbitrary is able to ', () => {
```
arestivo, line 238: This is a poor description for a test. 

`it('should be able to...'`

arestivo (commented):
Just address these last few comments and it's good to merge.

src/arbitraries/MappedArbitrary.ts
```diff
@@ -1,9 +1,16 @@
...
 export class MappedArbitrary<A, B> extends Arbitrary<B> {
-  constructor(public readonly baseArbitrary: Arbitrary<A>, public readonly f: (a: A) => B) {
+  constructor(
+    public readonly baseArbitrary: Arbitrary<A>,
+    public readonly f: (a: A) => B,
+    public readonly config?: MappedArbitraryConfig<A,B>
+  ) {
     super()
+
+    if (this.config && this.config.customCanGenerate)
```
ruippeixotog, line 12: nit: this can be a one-liner. 

```suggestion
    this.canGenerate = this.config?.customCanGenerate || this.canGenerate
```

src/arbitraries/types.ts
ruippeixotog, line 14: I am struggling to understand why you'll need `customCanGenerate`. I don't see it being used here or in #156 and if I understand correctly the meaning of the two configs here, `mappedArbitrary.canGenerate(value)` must be in all cases equal to `inverseFunction(value).some(v => baseArbitrary.canGenerate(v))` - no overrides can exist, except possible for optimized versions (which use cases?).

src/arbitraries/types.ts
```diff
@@ -8,3 +8,8 @@ export type ArbitrarySize = {
   type: 'exact' | 'estimated'
   credibleInterval?: [number, number]
 }
+
+export type MappedArbitraryConfig<A,B> = {
+  inverseFunction?: (b: B) => A | A[],
```
ruippeixotog, line 13: As mentioned above, return-type polymorphism is bad practice and in this case it won't deal correctly with arrays. You should either return always `A[]` or create different optional configs `injectiveInverseFunction` and `inverseFunction`.

ruippeixotog (changes requested):

ruippeixotog:
General comment: when you create stacked PRs like #153 and #156 - make sure to say clearly on the dependent one (#156) that it depends on another (#153). GitHub doesn't deal well with this I started reading it any only later realized that I was actually reviewing two PRs in the same place.

RuiGuedes:
@ruippeixotog Yeah I know sorry. I thought that PR #156 depended on this one and ended up merging the code from this one to the other but in the end, I verified that I was wrong. In fact, I commented on PR #156 saying that it depended on this one but ended up deleting such a comment. My mistake was not in reverting the merge.

src/arbitraries/MappedArbitrary.ts
RuiGuedes: Nice suggestion! Already considered it.

src/arbitraries/MappedArbitrary.ts
RuiGuedes: I know the problem is that for some reason that I could not understand I get a syntax error. I tried to research to find why it is giving me an error but I cannot find the reason for the problem.

```
        const inverseValues = !(this.config?.inverseFunction) ? [pick.original] :
                                            ^

SyntaxError: Unexpected token '.'
```

test/arbitrary.test.ts
RuiGuedes: I agree with you but it was already like this in the `master` branch. I already fixed the test description. Can you confirm that it is ok now?

src/arbitraries/types.ts
RuiGuedes: Already renamed the variables as you suggested!

src/arbitraries/MappedArbitrary.ts
RuiGuedes: I already figured it out. The problem was related with some problems with node and my ubuntu that was always resetting my node version to one that did not support using `?.`. Btw already made the change.

src/arbitraries/MappedArbitrary.ts
RuiGuedes: Yeah I know but I was getting a problem in doing that, that now it is fixed. Already added your suggestion!

src/arbitraries/types.ts
RuiGuedes: The decision of including both `inverseFunction` and `customCanGenerate` came up from some discussions that me, @arestivo, and @hugoferreira had. I will let them explain their point of view.

src/arbitraries/types.ts
RuiGuedes: I know that it is not frequent (at least not to me) to have return-type polymorphism but I didn't know that it was a bad practice. Therefore I will change to return always `A[]` because it is a more simple and less verbose solution.

RuiGuedes: Already added the modifications mentioned above.

src/arbitraries/MappedArbitrary.ts
RuiGuedes: I was doing that just to keep the code clean and because it was an easy way to achieve what I wanted. However, I decided to remove the return-type polymorphism (as suggested by @ruippeixotog), which turned out to remove the code that you mentioned here.

src/arbitraries/MappedArbitrary.ts
RuiGuedes: The code here being discussed has now changed (make sure you check the new modifications). Despite the alterations, I have added a new test using `ArbitraryArrays` and `MappedAribtrary` to make sure everything is working.

src/arbitraries/MappedArbitrary.ts
arestivo: Sometimes directly writing a _canGenerate_ function can be much easier than going the _inverse_ route.

src/arbitraries/types.ts
arestivo: I already had discussed this with Hugo and we completely agree with Rui Gonçalves.

src/arbitraries/types.ts
ruippeixotog: Copying @arestivo's comment here:

> Sometimes directly writing a _canGenerate_ function can be much easier than going the _inverse_ route.

That makes sense. Do we have any concrete examples of that to justify this more complex API? If not, it might be worth starting with `inverseFunction` and adding it later. Even if we choose to keep it, I would prefer if we disallowed setting _both_ functions (either using the type system or by asserting in runtime) as the issues resulting from inconsistent implementations may be very hard to debug.

arestivo: @ruippeixotog 

```
fc.integer().map(v => utils.sha1(v))
```

RuiGuedes: @ruippeixotog , @arestivo usual gives the example of finding the inverse function of `Math.random()` which is not possible. However, by specifying the `canGenerate` we know that if the value belongs to [0,1[, than it is possible to generate.

ruippeixotog: Sorry, I wasn't clear on the first comment - I was not asking for examples of non-invertible functions, of course :) I was looking for use cases that were meaningful for us as library developers (there are way more efficient ways to generate the space of all SHA1 and generating mapped arbitraries relying on `Math.random()` is a bad use of the library as randomness should be handled by FluentCheck itself).

I was initially assumed that the intention was to expose this to users, as this is the kind of fine-tuning that users don't generally care about (I never saw anything like this in other libraries, at least). As a user, I always saw shrinking as a nice to have but totally optional process for which I wouldn't make my tests more verbose. While getting 1 and "abc" as counterexamples is much nicer than 352853714817 or some long string with unicode chars, bad shrinking never prevented me from finding a root cause.

If you really prefer to expose this (i.e. announce it, not just having it as a hidden features), I'm ok with having both options assuming we assert that users don't provide both.

hugoferreira: There was extensive discussion about this, and I'm on the "no need for `canGenerate`" team. @arestivo convinced me that coming up with a `canGenerate` can be easier than coming up with an inverse function and I settled for: ok, let's have both. Now @ruippeixotog is convincing me again to move back to the "no need for `canGenerate`" team 😂

I _believe_ that this is not a function for the typical enduser for two reasons:

1. We maintain an history of `original` values with the sole purpose of _not_ needing an inverse when shrinking. We do _need_ a `canGenerate` but those are already provided by the internal arbitraries, including `map` and `filter`. Therefore, for the purpose of shrinking, the _user_ would never need to specify these as long as their arbitraries are a transformation of base arbitraries; do we all agree? 
2. The use case that led to this was the guided search being implemented in #146. We needed a way to find if a certain constant would be _ generateable_ by a certain arbitrary (e.g. `string`'s). But if they were being defined as `map`'s of `integer`'s (e.g. `string`'s), that would not be possible. But defining custom search strategies is not something the typical end-user would be doing...

So, uhhhh, yeah 😅 I would prefer to just have the inverse in map (it's more readable), but I know @arestivo has strong opinions about this 😂

ruippeixotog: > We do *need* a `canGenerate`

We need it only if we consider shrinking to be a requirement rather than a nice to have. It's good if we can provide it, but I'm perfectly ok with not having `canGenerate` for mapped and chained arbitraries built by the user - the added API complexity does not seem worth it.

> Therefore, for the purpose of shrinking, the user would never need to specify these as long as their arbitraries are a transformation of base arbitraries; do we all agree?

Not sure if I agree with this. If we consider shrinking to be a necessity I'm not sure how you'd try to shrink `exists('sha', fc.integer().map(v => utils.sha1(v)).then(sha => sha.endsWith("000"))` (a transformation of base arbitraries) just by looking at the history of values (maybe you can ask the crypto mining community... 🙂 ). If we agree it's a best effort then we really don't need it, but we *never* need it (no "as long as").

hugoferreira: > I'm perfectly ok with not having canGenerate for mapped and chained arbitraries built by the user

I'm also ok, but the need for `canGenerate` (the method, not the override @RuiGuedes provided) is for shrink to work with arbitraries _built by us_ (e.g. `string`), because they are internally defined as a sequence of `map`'s. We _need it_ to exist, although we don't need it to be _public_. Am I missing something?

> `exists('sha', fc.integer().map(v => utils.sha1(v)).then(sha => sha.endsWith("000"))`

I'm also not sure I agree with you. The example you provide seems to point to a limitation of the search process (finding an hash that ends in `000` is hard), rather than a counter-argument to the internal `canGenerate` definition that is used *after* we've found an example, and that it follows (as you say) a best effort. No?

ruippeixotog: > I'm also ok, but the need for `canGenerate` (the method, not the override @RuiGuedes provided) is for shrink to work with arbitraries _built by us_

Exactly 👍 And hence my question above: do we, library developers, have any use case where a custom `canGenerate` is feasible but `inverseFunction` is not?

> The example you provide seems to point to a limitation of the search process

Yeah, not the best example. I just wanted to illustrate that if we take into consideration everything that end users may try on top of our base arbitraries, you'll find cases where finding an example is easy but there's no obvious or meaningful way to shrink.

hugoferreira: > Exactly 👍 And hence my question above: do we, library developers, have any use case where a custom `canGenerate` is feasible but `inverseFunction` is not?

My initial answer to your question was _no_. So I'll let @arestivo chime in an present a defense on his perspective 😅

> Yeah, not the best example. I just wanted to illustrate that if we take into consideration everything that end users may try on top of our base arbitraries, you'll find cases where finding an example is easy but there's no obvious or meaningful way to shrink.

Point taken 👍 Once again, the inverse function solved our motivational use case (#146), and it was targeted for users that wanted to extend the library.

RuiGuedes: > The use case that led to this was the guided search being implemented in #146. We needed a way to find if a certain constant would be generateable by a certain arbitrary (e.g. string's). But if they were being defined as map's of integer's (e.g. string's), that wasn't possible. However, defining custom search strategies is not something the typical end-user would be doing...

Just to clear out, indeed the motivation behind the PR was PR #146 as @hugoferreira stated. However, later on, I decided, in PR #156 to enhance the string builder by implementing several other basic arbitraries. When I finished this last PR (#156) I realized that the modifications that the current PR introduce are no longer needed. The reason behind not needing them now is because since `strings` now use ASCII codes it is now possible for a given a string to compute their `original` (array of ASCII codes) value. Therefore the `canGenerate` method can be easily called with `value` and `original`. The way strings were implemented before PR #156 did not allow obtaining the original string value, hence the need for an inverse function.

Considering what I have just said, and considering your last conversation, I'm also on the side of having only the `inverseMap` function.

arestivo: Whatever! I could ask the exact same questions you have about the _inversemap_ method that you ask about the _canGenerate_ method. Let's merge this...

arestivo (approved):

hugoferreira:
@RuiGuedes I guess you have a lot of conflicts now... Can you please solve those?

RuiGuedes:
@hugoferreira Yeah I will solve them. Just give me a few minutes to finish something.

src/arbitraries/types.ts
RuiGuedes: @hugoferreira Already managed the conflicts with the `master` branch. However, I think we didn't entirely finish this conversation. How should we proceed xD?

hugoferreira: Let's settle meet in the middle and at least have the type system disallow specifying both, ok @ruippeixotog and @arestivo? The following code is untested, but I think we can change the constructor of the `MappedArbitrary` to:

```typescript
constructor (..., public readonly shrinkHelper?: { canGenerate: (pick: FluentPick<B>) => boolean} | { inverseMap: (b: B) => A[]}) 
```

arestivo: > > I'm ok with having only `inverseMap`.

RuiGuedes: Ok, I am on it. Tomorrow morning I will finish this PR and believe it can be merged then.

RuiGuedes: @hugoferreira Already addressed the proposed modifications on the last commit including adding some documentation to the `map` function. Feel free to check what I have done, especially the documentation of the `map` function xD. If everything is ok you can approve this and proceed with the merge.

hugoferreira:
I think this could be further simplified, but let's merge it for now so we don't get stuck and improve it later.

PR #159, opened 2021-03-17
Output test cases

### LINKED ISSUE #158: Output test cases
Test cases should be passed via FluentResult.

tiagoacastro:
Closes #158


src/FluentCheck.ts
```diff
@@ -8,7 +8,8 @@ type WrapFluentPick<T> = { [P in keyof T]: FluentPick<T[P]> }
 type FluentPicks = Record<string, FluentPick<any> | any>
 
 export class FluentResult {
-  constructor(public readonly satisfiable = false, public example: FluentPicks = {}) { }
+  constructor(public readonly satisfiable = false, public readonly example: FluentPicks = {},
+    public readonly testCases: Set<String> = new Set()) { }
```
hugoferreira, line 12: I don't think you should use a `Set`, otherwise you risk deduplicating _duplicated_ test cases. In fact, I think "(%) of duplicated tests" would be a great metric to report later.

hugoferreira (changes requested):

src/FluentCheck.ts
tiagoacastro: I agree, it can be an interesting metric

tiagoacastro:
For now, this information is only being presented to the user on a failed test, but later most of it will be presented also in case of a success

hugoferreira (changes requested):

tiagoacastro:
Here is the example of a test and its output on error:

163228821_495040228323747_1707100580033974356_n

163528903_189719742920438_2125571585289854817_n

162720436_2548965165406894_2221837449821341615_n

src/FluentCheck.ts
```diff
@@ -11,14 +11,16 @@ export class FluentResult {
   constructor(
     public readonly satisfiable = false,
     public example: FluentPicks = {},
-    public readonly seed?: number) { }
+    public readonly seed?: number,
+    public readonly withStatistics: boolean = false,
```
arestivo, line 15: What does showing test cases have to do with statistics?

src/FluentReporter.ts
```diff
@@ -10,11 +10,21 @@ export class FluentReporter extends Error {
...
-    msg.push('')
-    this.message = msg.join('\n')
+
+    if (result.withStatistics) {
+      msg.push('\n\nTest cases:\n')
+      msg.push(JSON.stringify(result.testCases))
+    }
+
+    msg.push('\n')
+    this.message = msg.join('')
```
arestivo, line 28: Not sure about the output format but we can look into that later.

src/arbitraries/types.ts
```diff
@@ -20,4 +20,6 @@ export class FluentRandomGenerator {
   }
 
   initialize() { this.generator = this.builder(this.seed) }
+
+  seeded() { return this.generator === Math.random ? undefined : this.seed }
```
arestivo, line 24: Is this from this PR?

arestivo (approved):
LGTM besides the name of the `withStatistics` variable but I'm fine with it if there's a reason behind it.

src/FluentCheck.ts
tiagoacastro: > What does showing test cases have to do with statistics?

My idea is that the flag will be used not only for showing test cases but also for all the more complex confidence improving aspects of the output like graphs and variable distributions that would increase the test execution time.

src/arbitraries/types.ts
tiagoacastro: > Is this from this PR?

no, it is not, I can remove it and create a new PR just for that

src/FluentCheck.ts
arestivo: Then call it `withTestCaseOutput` (or a better name) and when the user wants to turn on statistics (using `withStatistics`), the test case output collection is also turned on but without actually outputting in the end.

arestivo: To clarify:

`.withStatistics()` (should collect test cases but not output them)
`.withTestCaseOutputs()` (should collect test cases and output them)
`.withTestCaseOutputs().withStatistics()` (should collect test cases, output them and output statistics)

arestivo: I'm going to hijack this thread with some future ideas for this feature. For example: `.withTestCaseOutputs({onSuccess: false, maxTestCases: 100})`.

tiagoacastro:
> Here is the example of a test and its output on error:
> 
> 163228821_495040228323747_1707100580033974356_n
> 
> 163528903_189719742920438_2125571585289854817_n
> 
> 162720436_2548965165406894_2221837449821341615_n

Test outputing test cases on fail will now look like this (outputing certain information on success as well is a future issue)

image

hugoferreira:
> > Here is the example of a test and its output on error:

Can we please stop using screenshots unless it's not possible to show it as a copy-pastable snippet of text?

RuiGuedes:
@tiagoacastro @arestivo I am working on something that will mess up this PR way of doing things. By messing up I mean that most of the proposed changes will be discarded/modified. Therefore, in my opinion, this PR should depend on #170. However, if you feel that you still want to merge it on `master` feel free to do it, but notice that I will have to change it anyway.

src/FluentCheck.ts
```diff
@@ -56,7 +58,9 @@ export class FluentCheck<Rec extends ParentRec, ParentRec extends {}> {
 
   protected run(
     testCase: WrapFluentPick<Rec> | Rec,
-    callback: (arg: WrapFluentPick<Rec> | Rec) => FluentResult): FluentResult {
+    callback: (arg: WrapFluentPick<Rec> | Rec) => FluentResult,
+    testCases: ValueResult<any>[],
+    _partial: FluentResult | undefined = undefined): FluentResult {
```
hugoferreira, line 63: I'm 99% certain this `_partial` was removed. Why is it being reintroduced?

hugoferreira (changes requested):
We are missing tests for this PR. Existing tests are targeting generators. As they stand, if I change the `testCases` to always return `[]`, then those tests will pass.

src/FluentCheck.ts
tiagoacastro: yes, my bad, I didn't notice it while merging master (should've rebased but when I noticed it was too late).

tiagoacastro:
> We are missing tests for this PR. Existing tests are targeting generators. As they stand, if I change the `testCases` implementation to always return `[]`, those would still pass despite the clear out-of-spec behavior.

I fixed the previous tests that did not contain a then call and did not request for the testcases to be gathered, I also added a new test that verifies if two scenarios with different seeds generate different test cases which not only checks for empty array but also for other hardcoded values.

test/generation.test.ts
```diff
@@ -47,31 +47,49 @@ describe('Generation tests', () => {
...
+    const sc2 = fc.scenario()
+      .config(fc.strategy().defaultStrategy().withTestCaseOutput())
+      .withGenerator(prng, 4321)
+      .forall('a', fc.integer(-10, 10))
+      .forall('b', fc.integer(-10, 10))
+      .then(({a, b}) => a+b === b+a)
 
-      expect(sc1.check().testCases).to.eql(sc1.check().testCases)
-    */
+    expect(sc1.check().testCases).to.not.eql(sc2.check().testCases)
```
hugoferreira, line 93: This still doesn't constitute a specification for what we expect from testCases. Let's imagine that seed `1234` produced `[{a: 1000, b: 4000}]` and seed `4321` produced `[{a: 500, b: 100}]`. Both tests would still pass, and it is clearly not doing what we expect it to be doing.

A semantically rich test would check if the contents of `testCases` can be generated by the defined arbitraries. It would also if it doesn't contain more arbitraries than those given. It could also assert that it contains at least one element. And there might be other things that I'm simply forgetting...

Those testes should also not be part of the suite pertaining generation. I would advice to define them in a `testcases.test.ts`.

hugoferreira (changes requested):

test/generation.test.ts
tiagoacastro: I have pushed new changes containing the tests I believe encompass the behaviour of the test case output

PR #160, opened 2021-03-18
Minor safety issues with "truthy" operators

hugoferreira:
Replace two unsafe usages of "truthy" and || with explicit comparison with undefined and ??.

### DIFF
--- a/src/FluentCheck.ts
+++ b/src/FluentCheck.ts
@@ -178,7 +178,7 @@ abstract class FluentCheckQuantifier<K extends string, A, Rec extends ParentRec 
       }
     }
 
-    return partial || new FluentResult(!this.breakValue)
+    return partial ?? new FluentResult(!this.breakValue)
   }
 
   abstract breakValue: boolean
--- a/src/arbitraries/MappedArbitrary.ts
+++ b/src/arbitraries/MappedArbitrary.ts
@@ -13,7 +13,7 @@ export class MappedArbitrary<A, B> extends Arbitrary<B> {
 
   pick(): FluentPick<B> | undefined {
     const pick = this.baseArbitrary.pick()
-    return pick ? this.mapFluentPick(pick) : undefined
+    return pick !== undefined ? this.mapFluentPick(pick) : undefined
   }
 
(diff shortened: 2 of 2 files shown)


arestivo (approved):
LGTM

RuiGuedes (approved):

ruippeixotog:
LGTM, but why are you saying the usages are unsafe? Neither `FluentResult` nor `FluentPick<T>` can be falsy as they are a class and a non-empty object respectively. Is the intention to be resilient against casts that circumvent the type system? If that's the case I think we have a lot more to worry 😅

hugoferreira:
> LGTM, but why are you saying the usages are unsafe? 

You are 100% right.

For the case of objects, their usage do not constitute a problem. For other types, they do. Anyway, this was due to us finding a lot of `while(stuff !== undefined)` transformed into `while(stuff)` that although for objects it can work, it would be better to avoid using it as a guideline so that we don't introduce unintended bugs.

As a side note, I don't know about you guys, but my brain automatically infers a boolean whenever I see things in places I'm expecting them, such as `while ( _ )` and ` _ ? _  : _ ` 😅

ruippeixotog:
> that although for objects it is safe, it would be better to avoid using it as a guideline so that we don't introduce unintended bugs.

Let's use the linter for that then, instead of relying on everyone's code styles - it seems you can enable https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/strict-boolean-expressions.md.

(in fact, this is work for the type checker, not for a linter :( I was looking for a Typescript equivalent of Flow's [sketchy-null]( https://flow.org/en/docs/linting/rule-reference/#toc-sketchy-null), but I couldn't find anything...)

hugoferreira:
> Let's use the linter for that then, instead of relying on everyone's code styles - it seems you can enable https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/strict-boolean-expressions.md.

Uuuuuuh, I like that! I was just discussing with @arestivo that this should be the work of the linter 😅

PR #163, opened 2021-03-20
Present execution time of the test

### LINKED ISSUE #162: Present execution time of the test
The user should be informed of the execution time of the test.

tiagoacastro:
Closes #162


tiagoacastro:
The decision to use a package for fetching current time instead of using the Date was done based on [this thread](https://stackoverflow.com/a/15641427). Do you think a new dependency is worth it?

tiagoacastro:
Here is the output of a failed test

Screenshot from 2021-03-20 18-12-55

PR #169, opened 2021-03-25
Added nyc test coverage

hugoferreira:
Self-explanatory, and something that @RuiGuedes is playing with. The current output of `npm run coverage` is something like this:

```text
---------------------------|---------|----------|---------|---------|----------------------
File                       | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s
---------------------------|---------|----------|---------|---------|----------------------
All files                  |   81.69 |     60.8 |   77.76 |   84.64 |
 src                       |   79.82 |    82.98 |   72.73 |   81.18 |
  FluentCheck.ts           |   93.18 |    88.89 |   94.29 |   92.68 | 135-145
  FluentReporter.ts        |     100 |      100 |     100 |     100 |
  index.ts                 |     100 |      100 |      80 |     100 |
  statistics.ts            |   90.48 |    92.86 |   72.22 |   89.13 | 72,74-75,86-87
 src/arbitraries           |   67.82 |    54.21 |   60.66 |   68.48 |
  Arbitrary.ts             |   95.12 |    81.82 |      80 |   94.44 | 110,117
  ArbitraryArray.ts        |   96.97 |       75 |   93.33 |    96.3 | 58
  ArbitraryBoolean.ts      |    87.5 |        0 |      80 |   85.71 | 9
  ArbitraryComposite.ts    |      92 |        0 |    87.5 |   88.89 | 41-42
  ArbitraryConstant.ts     |   88.89 |        0 |   83.33 |   88.89 | 15
  ArbitraryInteger.ts      |   93.33 |    76.47 |   88.89 |   96.15 | 53
  ArbitraryReal.ts         |     100 |        0 |     100 |     100 | 4
  ArbitrarySet.ts          |   66.67 |     37.5 |   66.67 |      60 | 24-42,61
  ArbitraryString.ts       |       0 |        0 |       0 |       0 | 6-11
  ArbitraryTuple.ts        |   94.44 |    88.89 |   85.71 |   93.55 | 69-70
  BaseArbitrary.ts         |       0 |        0 |       0 |       0 | 8-53
  ChainedArbitrary.ts      |   83.33 |    33.33 |   71.43 |      80 | 21-25
  FilteredArbitrary.ts     |   88.89 |    57.14 |      90 |   94.74 | 54
  MappedArbitrary.ts       |   93.75 |    71.43 |      90 |   92.31 | 40
  NoArbitrary.ts           |   61.54 |        0 |      60 |   66.67 | 5,11,13-14
[...]

PR #170, opened 2021-03-25
Add coverage measurement on each test run

RuiGuedes:
In order to implement `coverage-guided` strategies, we need to measure coverage each time a given test runs with a particular set of inputs. Coverage should be tracked so that a relation between coverage and inputs can be established. More details will be given as the development advances.


RuiGuedes:
Depends on #146 and #169.

RuiGuedes:
This PR is ready for review. It is worth mentioning that this PR purpose is to measure coverage on each test run and not making the input generation be guided by coverage. This last objective will be achieved in a future PR.

PR #177, opened 2021-03-30
Constant extraction buffer overflow detection

RuiGuedes:
This PR aims to enhance the current state of the feature responsible for extracting constants from the code. The enhancement consists of creating strings whose size matches the numerics extracted from the code. 

A new test, derived from a buffer overflow [example](https://gist.github.com/RuiGuedes/42dc9482a6de322301ac40adedc52fb7), was created to show the importance of this enhancement.

### Test

```
it('should be able to build and use strings with length matching the numerics extracted', () => {
      expect(fc.scenario()
        .config(fc.strategy()
          .withRandomSampling(100)
          .withBias()
          .withConstantExtraction()
        )
        .forall('input', fc.string(0, 64))
        .then(({input}) => {
          if (input.length >= 20) return false
          return true
        })
        .check()
      ).to.have.property('satisfiable', false)
    })
```


RuiGuedes:
Depends on #170

PR #178, opened 2021-04-01
Implementation of nwise and pairwise combinators

RuiGuedes:
This PR aims to create `nwise` and `pairwise` combinator functions. Such functions will be used in the future to group the randomly generated test cases in pairs (to increase efficiency) and to create the seed collection (composed by `corner cases` and `extracted constants`) that will be used by coverage-guided strategies.

PR #179, opened 2021-04-02
Combining random sampling with pairwise testing

RuiGuedes:
This PR aims to refactor the current state of fluent check test cases' generation process to include the possibility of executing pairwise testing. The major goal behind this PR is to change the codebase in order to be ready for the integration of the coverage-guided strategy.


RuiGuedes:
Depends on #170, #177, and #178

RuiGuedes:
This PR is ready for review. Since it contains a huge refactor on the FluentCheck's testing process, it should only be reviewed when the PRs which this one depends on are fully merged. 

In this PR I have also tried to make the workflow as fully optimized as possible. Therefore all new possible optimizations should be done in a new PR since this one purpose was to introduce the possibility of choosing between complete-wise and pair-wise testing.

Note: Some of the developed code looks that it can be refactored to use javascript built-in functions but the reason why such functions are not being used is that they can compromise performance as I have verified that during profiling process.

PR #189, opened 2021-04-17
Adding timeout option to stop the testing process

RuiGuedes:
This PR aims to introduce the possibility to stop testing not only when a property is verified/falsified or when there are no more inputs to be tested but also when a certain timeout has been reached. This way we can make the framework more extensible. This feature will be more crucial when the testing process is coverage-guided based.


RuiGuedes:
Depends on #170, #177, #178, and #179

PR #190, opened 2021-04-19
Adding stopping criterion based on coverage levels

RuiGuedes:
This PR aims to introduce the possibility to stop testing once a certain level of coverage is achieved. Since there is no possibility of making sure that a certain level of coverage is achieved this proposal will have as a secondary stop criterion mechanism the timeout option developed in #189.


RuiGuedes:
Depends on #170, #177, #178, #179, and #189

PR #191, opened 2021-04-19
Implementation of coverage-guided fuzzing

RuiGuedes:
This PR aims to introduce coverage-guided fuzzing. So far, coverage is already being measured and used as one of the testing process stop criteria. The idea in this PR is to implement all the logic typically associated with coverage-guided fuzzing (seed inputs, mutations, etc). The overall logic will be based on the following algorithm, which is based on [Zest](https://rohan.padhye.org/files/zest-issta19.pdf) and [AFL](https://github.com/google/AFL): 

```
Input: program p
Output: a set of test inputs and failing inputs
---
S ← {bias + constants}
F ← ∅
totalCoverage ← ∅
repeat
   for param in S do
      for 1 ≤ i ≤ numCandidates(param) do
         coverage, result ← run(p, candidate )
         if result = Failure then
            F ← F ∪ candidate
         else if coverage ⊈ totalCoverage then
            S ← S ∪ {candidate}
            totalCoverage ← totalCoverage ∪ coverage
until given timeout or minimum coverage is achieved
return F
```


RuiGuedes:
This PR is ready for review. If you reviewed this PR carefully, you would see that the logic behind the mutations of each arbitrary is not yet implemented. In fact, I left a _**TODO**_ in the `Arbitrary` `mutate()` function. 

That makes the implementation of coverage-guided fuzzing incomplete because mutations are not happening as they should. However, that's on purpose because the mutations' logic will be implemented later on another PR since it encompasses complex changes.

PR #194, opened 2021-04-21
Implemented Arbitrary abstract mutate function

RuiGuedes:
This PR aims to implement the `Arbitrary` abstract `mutate()` function in order to allow the coverage-guided strategy to properly mutate the inputs used in the testing process.

## Arbitraries

- [x] ArbitraryArray
- [x] ArbitraryBoolean
- [x] ArbitraryComposite
- [x] ArbitraryConstant
- [x] ArbitraryInteger
- [x] ArbitraryReal
- [x] ArbitrarySet
- [x] ArbitraryTuple
- [x] ChainedArbitrary
- [x] FilteredArbitrary
- [x] MappedArbitrary
- [x] NoArbitrary
- [x] WrappedArbitrary

PR #197, opened 2021-04-25
Constants extraction process improvement

RuiGuedes:
This PR aims to improve the constants extraction process. The improvement consists of enhancing the integers extraction by applying basic arithmetic operations on the extracted constants.

I also thought that the `canGenerate` verification of extracted constants could be removed so that constants could be extracted and used regardless of being possible to generate by a given arbitrary or not. However, after deeper analysis, I verified that it does not make sense because such a situation would only benefit the overall testing process if the defined test was wrongly defined. Therefore, I believe that the presence of the `canGenerate` method makes sense and therefore I decided to keep it as it is.


RuiGuedes:
Depends on #194

Issue #427, opened 2025-11-27
Add ArbitraryRecord for generating structured objects


hugoferreira:
# Change: Add ArbitraryRecord

## Why

Property-based tests frequently need to generate structured objects with named properties. Currently, users must either:
1. Use `tuple()` and manually destructure positional values
2. Use `given()` chains to incrementally build objects
3. Manually compose arbitraries with `map()`

This creates verbose, error-prone code when testing domain objects, API payloads, or configuration structures. A `record()` combinator would provide a direct, type-safe way to generate objects from a schema of arbitraries.

## What Changes

- Add new `ArbitraryRecord<S>` class that generates objects from a schema
- Add `fc.record(schema)` factory function in arbitraries index
- Support full type inference from schema to output object type
- Implement proper shrinking (shrink each property independently)
- Implement corner cases as combinations of property corner cases
- Export new arbitrary from main index

## Impact

- Affected specs: `arbitraries`
- Affected code:
  - `src/arbitraries/ArbitraryRecord.ts` (new)
  - `src/arbitraries/index.ts` (add export)
  - `src/arbitraries/internal.ts` (add export)
  - `test/record.test.ts` (new)

Issue #30, opened 2020-06-22
Support more out-of-the box arbitraries


hugoferreira:
We currently provide arbitraries for:

## Primitives
- [x] Integers
- [x] Natural numbers
- [x] Reals
- [x] Booleans
- [x] Strings

## Collections
- [x] Arrays
- [x] Sets
- [ ] Dictionaries (`Maps`? `Objects`?)
- [x] Tuples

## Special cases
- [x] Constants

PR #428, opened 2025-11-27
feat: add ArbitraryRecord for generating structured objects

hugoferreira:
## Summary

- Add `fc.record(schema)` function for generating objects from a schema where keys map to arbitraries
- Full TypeScript type inference from schema to output object type
- Proper shrinking (shrinks each property independently)
- Corner case generation (combinations of property corner cases)

**Proposal:** openspec/changes/add-arbitrary-record/proposal.md
**Closes:** #427, #30

## Test Plan

- [x] All existing tests pass (277 tests)
- [x] New tests added in `test/record.test.ts` covering:
  - Basic record generation with primitive arbitraries
  - Nested records
  - Empty schema handling
  - NoArbitrary in schema
  - Size calculation
  - Corner cases generation
  - canGenerate validation
  - Shrinking behavior
  - Property-based tests
  - Type inference verification
- [x] `openspec validate add-arbitrary-record --strict` passes
- [x] `npm run lint` passes


arestivo (approved):
LGTM

Issue #481, opened 2025-12-02
refactor: adopt array methods over manual iteration


hugoferreira:
# Change: Adopt Array Methods Over Manual Iteration

> **GitHub Issue:** (to be created)

## Why

After enabling `noUncheckedIndexedAccess: true`, many `for...in` loops were added with explicit undefined checks. However, array methods like `every()`, `some()`, `filter()`, `slice()`, and `at()` handle undefined naturally and are more idiomatic, reducing code noise and improving readability.

**Current Problem:**
- `for...in` loops with `Number()` conversion and explicit undefined checks
- Manual iteration with verbose undefined handling
- Less declarative code that's harder to read and maintain
- Repeated patterns that could use built-in array methods

**Solution:**
Replace `for...in` loops and manual iteration with declarative array methods that handle undefined naturally and work well with type guards.

### Array Methods Pattern

1. **`every()` and `some()`** - For validation/checking operations:
   - Replace `for...in` loops that check all/some elements
   - Handle undefined naturally in predicates
   - More declarative and readable

2. **`filter()` with Inferred Type Predicates (TypeScript 5.5+)** - For filtering undefined values:
   - **TypeScript 5.5 automatically infers type predicates** - `filter(item => item !== undefined)` returns `NonNullable<T>[]` without explicit type guard!
   - No need for explicit `(item): item is NonNullable<T>` - TypeScript infers it
   - Return `NonNullable<T>[]` types automatically
   - Eliminate manual undefined checks and type assertions

3. **`slice()`** - For bounds-safe array access:
   - Automatically handles out-of-bounds indices
   - Returns empty array for invalid ranges
   - Eliminates manual bounds checking

4. **`at()` with Nullish Coalescing** - For safe indexed access:
   - Returns `T | undefined` naturally
   - Use `??` for defaults
   - More explicit than manual index access

5. **`findIndex()` and `find()`** - For searching operations:
   - Handle undefined naturally
   - Return `-1` or `undefined` for not found
   - More idiomatic than manual loops

### Code Analysis

**Current Issues Found:**

1. **`src/arbitraries/ArbitraryTuple.ts:66-70`** - `for...in` loop with manual checks:
   ```typescript
   for (const i in value) {
     const index = Number(i)
     if (!this.arbitraries[index].canGenerate({value: value[index], original: original[index]}))
       return false
   }
   ```
   **Problem:** `for...in` with `Number()` conversion, no bounds validation, verbose
[...]

PR #490, opened 2025-12-02
refactor: adopt array methods over manual iteration

hugoferreira:
## Summary

Implements openspec change proposal.

**Proposal:** openspec/changes/refactor-strict-mode-array-methods/proposal.md
**Closes:** #481

## Tasks

See `openspec/changes/refactor-strict-mode-array-methods/tasks.md` for implementation checklist.

## Test Plan

- [ ] All existing tests pass
- [ ] New tests added for changed behavior
- [ ] `openspec validate refactor-strict-mode-array-methods --strict` passes

### DIFF
--- a/openspec/changes/refactor-strict-mode-array-methods/tasks.md
+++ b/openspec/changes/refactor-strict-mode-array-methods/tasks.md
@@ -1,59 +1,59 @@
 ## 1. Code Analysis and Cataloging
 
-- [ ] 1.1 Search for `for...in` loops with array index access
-- [ ] 1.2 Search for manual array iteration with undefined checks
-- [ ] 1.3 Identify loops that can use `every()`, `some()`, `filter()`
-- [ ] 1.4 Identify array access that can use `slice()` or `at()`
-- [ ] 1.5 Catalog all locations where array methods can replace manual iteration
+- [x] 1.1 Search for `for...in` loops with array index access
+- [x] 1.2 Search for manual array iteration with undefined checks
+- [x] 1.3 Identify loops that can use `every()`, `some()`, `filter()`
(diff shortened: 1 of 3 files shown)


hugoferreira:
Quick note on the `ArbitrarySet.cornerCases` refactor, since it looks like a semantic change at first glance.

The original implementation built `min`/`max` using indexed loops over `this.elements` with explicit `!== undefined` checks:

- Those loops were *not* searching for numeric / ordered minima or maxima of `A`.
- They were simply taking the first `min` and first `max` elements from `this.elements` to form representative sets of the smallest and largest *cardinality*.
- The tests codify this: `fc.set(['a', 'b', 'c'], 1, 3).cornerCases()` is expected to yield `[['a'], ['a', 'b', 'c']]` (see `test/arbitrary.test.ts`).

Given the constructor invariants:
- `this.max = Math.min(max, elements.length)` in `ArbitrarySet`, and
- `fc.set` only constructs `ArbitrarySet` when `min <= elements.length`,

we never actually index out of bounds in those loops, so the `!== undefined` checks were only there to satisfy `noUncheckedIndexedAccess`.

The new implementation:

- Uses `this.elements.slice(0, this.min)` and `this.elements.slice(0, this.max)` instead of manual loops.
- Keeps the exact same observable behaviour (same first-`min` and first-`max` subsets, same tests passing).
- Just removes redundant undefined checks and makes the intent "take the first N elements from the canonical domain array" explicit and declarative.

So the semantics remain: `ArbitrarySet.cornerCases` exposes representative **smallest** and **largest** allowed sets by *cardinality* from the underlying element domain, not value-based min/max over `A`.

PR #545, opened 2025-12-31
feat: CI calibration advanced studies fixes and enhancements

hugoferreira:
This PR implements comprehensive fixes and enhancements to the CI calibration advanced studies (#541), addressing critical methodology issues and adding statistical rigor.

## Key Changes

- **New Evidence Studies**: Added beta-prior-comparison, correlation-effects, and shrinking-ci-calibration studies
- **Updated Studies**: Enhanced adversarial-patterns, ci-calibration, ci-convergence, composition-depth, and early-termination studies
- **Analysis Infrastructure**: Added comprehensive Python analysis scripts with statistical validation
- **Documentation**: Updated evidence documentation with detailed findings and methodology
- **Worker Infrastructure**: Implemented parallel study execution capabilities
- **OpenSpec Updates**: Added detailed design specifications and task breakdowns

## Improvements

- Fixed methodology issues in all studies
- Added statistical rigor with proper power analysis
- Generated updated figures and raw data for all studies
- Enhanced reproducibility with deterministic study designs
- Comprehensive documentation of findings and recommendations

## Related

Addresses issues identified in the CI calibration advanced studies review (#541).


docs/evidence/ci-calibration-advanced.md
```diff
@@ -1,110 +1,163 @@
...
 
-![Weighted Union](figures/weighted-union.png)
+**Analysis**:
+- **Warm Start fails catastrophically** when shrinking changes the pass rate. In the `subset` scenario, the parent learned "50% pass rate" with high confidence. The child (now restricted to the valid region) has a 100% pass rate. The transferred posterior makes the child **confidently wrong** (estimating 50% size), leading to 0% coverage.
+- **Cold Start is robust**: By resetting the posterior (Scale 0), the child quickly relearns the new pass rate (100%), maintaining perfect calibration.
+- **Efficiency Tradeoff**: While Warm Start improved precision by ~50% in the constant-pass-rate scenario, the risk of massive miscalibration in other cases is unacceptable.
 
-## Key Conclusions
+**Conclusion**:
+**Do NOT implement Warm Start shrinking.** The current Cold Start approach (Scale 0) is the correct design choice for robustness. The "memory" of the parent arbitrary is harmful when the shrinking process biases the sample space (which it fundamentally does).
```
hugoferreira, line 142: ## How I’m going to review this

1. Read what the study *actually* does (not the narrative).
2. Check whether the metrics match the question (“should we transfer posterior?”).
3. Identify failure modes that are artifacts of the experimental setup vs intrinsic.
4. Propose concrete design alternatives and follow-up experiments.

---

## What the study *actually* measures (important nuance)

From the code, each trial does:

* Build a filtered integer arbitrary on `0..100`.
* “Warm up” the **parent** by calling `parentArb.pick(generator)` 50 times.
* Manually call `parentArb.shrink(pick)` where `pick.value` is fixed (25 or 50), to obtain a **single child arbitrary**.
* Immediately call `shrunkArb.size()` and record:

  * `estimatedSize`
  * the credible interval if `sizeInfo.type === 'estimated'`
  * whether a *hard-coded* `trueSize` lies within that interval. 

So this is primarily testing **initial calibration of the child’s size estimate right after a shrink step**, under different posterior-transfer scales (0, 0.5, 1.0). It is *not* testing multi-step shrinking dynamics, nor the child’s ability to “recover” after more sampling.

---

## Things that look correct in your interpretation

### Why the catastrophic failure is plausible

In the “subset” setup, the parent acceptance rate is ~50% (numbers `< 50` in `0..100`), while the child region is constructed to be entirely within the passing region (acceptance ~100%). Transferring a high-confidence Beta posterior from the parent will produce a child posterior that is **strongly concentrated around ~0.5**, and unless the child collects enough contrary evidence, its credible interval will be narrow and wrong → **systematic undercoverage**. That mechanism matches the observed “confidently wrong” story.

### Cold start robustness (for this specific measurement)

Because scale 0 discards the parent posterior, the child’s posterior is dominated by child evidence, which in the subset case is consistent (near 100% accept), so it calibrates quickly (again: *for the “immediate size() after shrink” measurement*).

---

## Gaps / potential methodological issues

### 1) “True size” is assumed, not computed from the child arbitrary

The study hard-codes `trueSize = 25` in both scenarios, based on a comment-level assumption about how the shrink constrains the integer space (“Integer shrinks to target-1”).

That’s brittle. If the actual shrink semantics are *slightly* different (inclusive vs exclusive bounds, or a different shrink domain than you described), then the *coverage calculation is invalid* (it might still show failure, but the magnitude and even direction can change).

**Actionable fix**

* In the trial, compute `trueSize` by enumerating the child space (these domains are tiny) or by an internal “exact size” path if available. Concretely: add a helper that exhaustively iterates `shrunkArb`’s underlying base domain (or uses any internal introspection hooks) and counts values satisfying the predicate, rather than assuming 25.

### 2) You’re only measuring the child **right after shrink**, not after any child learning

The child’s `size()` is measured immediately after creating the shrunk arbitrary. There is no controlled “child evidence budget” (e.g., run N picks on the child, then measure).

This matters because the design question is really:

> “Does transferring posterior help overall shrinking efficiency *without* harming correctness?”

If the system naturally performs multiple picks/shrink steps on the child before relying heavily on `size()`, then “initial miscalibration” might be transient and non-fatal.

**Actionable fix**

* Add a second phase: after `shrunkArb` is created, run `childBudget ∈ {0, 5, 20, 100}` child picks (or “attempts”, if you can count rejections), then evaluate calibration/width. Plot recovery curves.

### 3) Parent warm-up is fixed (50), which implicitly fixes parent posterior strength

Warm start failure severity depends heavily on the **effective sample size** of the transferred posterior (how concentrated the Beta is). Here, parent warmup is fixed at 50 accepted picks; depending on how the filter updates the posterior (counting attempts vs accepts), that may correspond to a much larger effective n.

**Actionable fix**

* Sweep parent warm-up: `parentBudget ∈ {5, 20, 50, 200, 1000}` and quantify the boundary where warm start turns from helpful to harmful.

### 4) Only two scenarios; one is maximally adversarial

The “subset” scenario is effectively a distribution shift from p≈0.5 to p≈1.0. That’s a good red-team case, but it’s not informative about *typical* shifts in real shrink traces.

**Actionable fix**

* Add intermediate shifts: p_child ∈ {0.6, 0.7, 0.8, 0.9} using synthetic predicates.
* Add the opposite shift (acceptance drops), which is another realistic failure mode.

### 5) Coverage is reported, but not compared to the nominal CI level

The study reports “coverage = 100%” as “calibrated” in some rows, but without stating the credible interval level (e.g., 90/95/99%) and without an uncertainty band on the empirical coverage estimate.

Even if CI level were 95%, observed 100% over finite trials can indicate conservatism (maybe acceptable, maybe not), but you need to interpret it statistically.

**Actionable fix**

* Report the nominal CI level alongside results.
* Add a confidence interval around empirical coverage (Wilson interval works well here).
* Explicitly test “coverage ≥ nominal − ε” (non-inferiority framing) instead of eyeballing.

---

## I would soften the “Do NOT implement warm start” conclusion (but keep the warning)

What your study *strongly* supports is:

> **Naïvely transferring the parent’s global acceptance-rate posterior into the child (especially at high effective strength) is unsafe under distribution shift.**

That’s a solid negative result.

But “do not implement warm start at all” is broader than what was tested. There are well-known robust Bayesian tricks that can preserve most of the benefit in the “scaled” case while bounding harm under shift.

---

## Actionable design alternatives that are likely to work

### A) Cap the effective sample size of the transferred posterior (simple, high value)

Right now, `scale` multiplies whatever concentration the parent has. If the parent posterior corresponds to (say) hundreds of effective trials, even `scale=0.5` can be too confident.

**Implement**

* When transferring `(α, β)` to child, rescale so that `(α+β)` is clamped:

  * `α' = 1 + min(Ncap, scale*(α-1 + β-1)) * (α-1)/(α-1+β-1)`
  * `β' = 1 + min(Ncap, scale*(α-1 + β-1)) * (β-1)/(α-1+β-1)`
* Start with `Ncap ∈ {5, 10, 20}`.

This turns warm start into “a small prior hint”, preventing “confidently wrong” behavior.

### B) Mixture prior (robust to shift)

Instead of transferring a single Beta, use a mixture:

* `π * Beta(parent) + (1-π) * Beta(base)` (base often uniform Beta(1,1))

This prevents the posterior from collapsing too hard around the parent’s belief.

**Implement**

* Approximate mixture by adding pseudo-counts from both components (or keep an explicit mixture if you already have machinery).

### C) Shift detection with an early reset (“trust, but verify”)

After the child observes a small number of filter attempts, test whether those outcomes are plausible under the transferred prior predictive.

**Implement**

* After `n0` child trials, compute a prior-predictive p-value (or Bayes factor) for observing `k` successes.
* If implausible (e.g., p < 1e-3), reset to cold start (or drop `π`, or reduce `Ncap` further).

This directly targets the failure mode your “subset” scenario exposed.

### D) “Spatial warm start” is promising, but you’ll need to define what “space” means

Your future-work note mentions IQR/kNN. That only becomes well-defined if:

* you have an ordered metric space (integers: yes; structured values: maybe no),
* and the filter acceptance has locality.

It’s worth exploring, but I’d treat it as a second-phase project after you implement A–C, because A–C are cheap and cover most of the risk.

---

## Concrete follow-up experiments (minimal effort, maximal clarity)

1. **Compute `trueSize` by enumeration** of the child space (remove the hard-coded 25).
2. Add **child evidence budgets** and plot calibration-vs-budget curves for each scale.
3. Sweep **parentBudget** and show where warm start breaks.
4. Test **capped transfer** (`Ncap`) + optional **shift reset** against your two scenarios:

   * Expectation: keep most of the “scaled” precision gain, eliminate 0% coverage in “subset”.

hugoferreira: @gemini-code-assist WDYT?

hugoferreira: @gemini-code-assist well, thank ChatGPT 5.2 thinking for that

hugoferreira: @gemini-code-assist This review plan shows a deep grasp of the system and correctly identifies the core risk: transferring a high-confidence posterior under distribution shift can cause systematic undercoverage.

However, after reviewing the implementation of `FilteredArbitrary.ts`, I’ve found a key confounding variable missing from the current review, along with some suggestions to refine your proposed methodology.

---

### 1. Critical Finding: The Hidden Warm-Up

The review assumes that the "Cold Start" baseline (scale 0) begins with a clean Beta(2,1) prior. This is **incorrect**.

The `FilteredArbitrary` constructor always performs a deterministic warm-up of 10 samples:

```typescript
// src/arbitraries/FilteredArbitrary.ts
const WARMUP_SAMPLES = 10;
let seed = 0xCAFEBABE;
for (let i = 0; i < WARMUP_SAMPLES; i++) {
  this.pick(lcg);
}
```

**Implications:**
- **Cold Start is actually "10-sample Warm Start":** In the "subset" scenario (child acceptance rate near 100%), these 10 samples will typically all pass, making the child's posterior start with roughly $\alpha \approx 12, \beta \approx 1$. This explains the strong baseline performance—it's "peeking" at the truth before measurements even begin.
- **Warm Start Failure Is Underestimated:** The fact that Warm Start can still fail even with these 10 perfect samples counterbalancing the transferred prior demonstrates how risky the transferred prior is: it overwhelms even new, accurate data.

---

### 2. Critique of "Actionable Fixes"

- **Dynamic `trueSize` (Agreed):** Avoid hardcoding values (e.g., 25). Implement a helper such as `countMatches(arbitrary, predicate)` for exhaustive enumeration in small domains.
- **Child Evidence Budget (Refined):** Rather than simply adding a sample budget, we must disable the constructor warm-up in the study to accurately compare "Prior Transfer" with a true "Cold Start" (scale 0, zero warm-up). Right now, "Prior Transfer" and "Constructor Warm-up" are always convoluted.
- **Capped Transfer (Agreed):** This is the correct path forward. Implement a `TransferStrategy` interface or static configuration within `FilteredArbitrary` to enable linear-scale, capped, and mixture strategies without hardcoding.

---

### 3. Refined Plan

To address these findings, I propose the following plan:

1. **Refactor `FilteredArbitrary` for Scientific Comparison:**
    - Add a static flag `FilteredArbitrary.ENABLE_CONSTRUCTOR_WARMUP` (default: `true`) to allow the study to disable warm-up as needed.
    - Refactor `WARM_START_SCALE` into a flexible `transferStrategy` method or callback, allowing easy testing of "Capped" and "Mixture" strategies.

2. **Rewrite the Study (`warm-start-shrinking.study.ts`):**
    - Disable constructor warm-up for all experimental trials.
    - Use dynamic truth: implement `getTrueSize()` via enumeration.
    - Parameter sweeps:
        - `parentBudget`: [10, 50, 200] (tests prior strength)
        - `childBudget`: [0, 5, 20] (tests recovery speed)
        - `scale`: [0, 0.5, 1.0], plus a new "Capped(N=10)" strategy

3. **Run & Analyze:** These changes will offer clear, decisive evidence to guide the final choice of shrinking strategy.

hugoferreira: 🤦‍♂️

